PR #5, opened 2016-12-24
Parameterize airtable base URL so that devs can point to alternate dataset

krish4good:
This allows for people to point to a personal/ test data set instead of the
production one. The production URL is still the default.


nickoneill (approved):
👍

nickoneill:
Looks great! I can copy a version of the airtable database we're working with if you need that as well

PR #6, opened 2016-12-31
Backend improvements

krish4good:
As described in the commits:

#### commit 1: Refactor airtable data access and implement refresh

* use the airtable-go MIT licensed library. This handles full CRUD as well
  as pagination for lists.
* Make the Contact and Issue objects the return value for getting all issues
* Wrap the airtable client with an in-memory cache that automatically refreshes
  from Db every so often.
* Make the minimum changes to the handlers for the refactored interface

#### commit 2 : Refactor civics API and use a new model object as return value

* Simplify civics API implementation to filter by role rather than level
* Create a new LocalReps model object and have the civics API return it
* API now also returns the normalized address as returned by Google
* Create a cache wrapper to decouple cache logic from webservice logic

#### commit 3 : Refactor handlers, lose global variables

* Extract handler functions into struct methods
* Handlers use interfaces to get their job done
* Routing simplified to mount GET and POST routes separately
* CORS handling moved to main

### DIFF
--- a/go/admin.go
+++ b/go/admin.go
@@ -4,18 +4,20 @@ import (
 	"net/http"
 )
 
-func adminHandler(w http.ResponseWriter, r *http.Request) {
-	if r.Method != "GET" {
-		http.Error(w, "GET only", 403)
-		return
-	}
-
+type cacheReloader interface {
+	Reload()
 }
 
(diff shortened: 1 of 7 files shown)


nickoneill:
👍

nickoneill:
Hard to make notes on these individually since everything is jumbled up. I'll send you some questions separately @krish4good

Issue #9, opened 2017-01-25
Mobile view could use some love


nickoneill:
- [x] Styling is not great on small screens
- [x] clicking result buttons doesn't let you know there's a new contact
- [x] phone numbers don't tap-to-dial

ivanoats:
I'm taking a look at the tap-to-dial issue now https://mobiforge.com/design-development/mobile-friendliness-101-adding-click-to-call-links-to-a-website

ivanoats:
Thanks, Zach 👍

nickoneill:
All set here for now

PR #11, opened 2017-01-25
Add phone link, scale font size

ZachGawlik:
Keeping my first open-source contribution small. 🙈  Checks off task on #9 

Phone before:
screen shot 2017-01-25 at 4 31 50 pm

Phone after
screen shot 2017-01-25 at 4 31 54 pm

Tablet before:
screen shot 2017-01-25 at 4 32 16 pm

Tablet after:
screen shot 2017-01-25 at 4 32 10 pm

### DIFF
--- a/static/js/components/contact.js
+++ b/static/js/components/contact.js
@@ -15,7 +15,11 @@ module.exports = (c, state, prev, send) => {
         <div class="call__contact__image"><div class="crop"><img src="${photoURL}"/></div></div>
         <p class="call__contact__type">Call this office:</p>
         <p class="call__contact__name">${c.name} ${repID}</p>
-        <p class="call__contact__phone">${c.phone}</p>
+        <p class="call__contact__phone">
+          <a href="tel:${c.phone}">
+            ${c.phone}
+          </a>
+        </p>
         <p class="call__contact__reason"><strong>Why you're calling this office:</strong> ${reason}</p>
       </div>
 	`;
--- a/static/scss/_call.scss
+++ b/static/scss/_call.scss
@@ -42,8 +42,12 @@
 
     &__phone {
       margin-bottom: 0.1rem;
-      font-size: 5rem;
+      font-size: 2.8rem;
       font-weight: 600;
+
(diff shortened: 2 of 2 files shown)


nickoneill:
Blue phone link feels a bit out of place here. I wonder if people will still tap it if it's not highlighted?

ZachGawlik:
@nickoneill I think there needs to be some visual signifier. I personally would not think to tap the black text phone numbers in the before photos. 

The link style echoes that of the "Tweet this issue", "Share this issue", and the "Change" button so it makes sense to reuse the style that signifies "you can click/tap this bit of text". Does it look less out of place to you when viewed within the rest of the screen?

screen shot 2017-01-25 at 8 27 15 pm

nickoneill:
I actually like the phone font size on the larger screen sizes, it really emphasizes the goal. Dial the number. Happy to merge the telephone links in now, but we need to tweak the actual number font sizes to be more appropriate for each level of screen size.

thejbru:
I see that I should have looked before sending my pull request. Very similar styling. I'd suggest adding padding to the link at the smaller sizes in order to make the tap target a little larger. I went with 2rem size in my edit, but if you're looking for more emphasis on the number I think 3rem works pretty well for most recent phones.

imaflamingo:
I was looking to add this feature before making some calls on my phone so I'm glad there's already a PR for it

imaflamingo:
When will this be merged? It's a great feature and the smaller phone font looks much better

nickoneill:
I just pushed tappable call links to the site - thanks for your help with that!

For the phone number size, it would be great if it could adjust size so it always fills the container it's in. That way desktop users can still have a large call to action and mobile users are more comfortable on the smaller screens.

thejbru:
The phone number size could be defined using the "vw" unit. That would scale it to the visual width of the device. Supported pretty well: http://caniuse.com/#search=vw

nickoneill:
This is superseded by the updated design changes - thanks for all the input!

PR #14, opened 2017-01-26
Responsive issues

ZachGawlik:
Working towards #9 by fixing the layout issues that stuck out to me when I very first visited the site.  I was on a ~1200px window on my laptop, so these were issues with the medium range

Stat before and after:
stat-before-small stat-after-small

Outcome buttons text overflow & footer position before:
screen shot 2017-01-25 at 7 35 47 pm

Outcome buttons & footer after:
screen shot 2017-01-25 at 7 35 54 pm

Outcome buttons on mobile before and after:
mobile-buttons-before mobile-buttons-after

### DIFF
--- a/static/js/components/call.js
+++ b/static/js/components/call.js
@@ -25,7 +25,7 @@ module.exports = (state, prev, send) => {
     if (result == null) {
       send('skipCall', { issueid: issue.id });
     } else {
-      send('callComplete', { result: result, contactid: currentContact.id, issueid: issue.id });      
+      send('callComplete', { result: result, contactid: currentContact.id, issueid: issue.id });
     }
   }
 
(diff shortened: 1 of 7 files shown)


static/js/components/call.js
```diff
@@ -43,17 +43,19 @@ module.exports = (state, prev, send) => {
       <div class="call__script__body">${issue.script.split('\n').map((line) => scriptLine(line, state, prev, send))}</div>
     </div>
 
-    <menu class="call__outcomes">
+    <div class="call__outcomes">
       <h3 class="call__outcomes__header">Enter your call result to get the next call:</h3>
-      <menuitem onclick=${() => outcome('unavailable')}>Unavailable</menuitem>
```
ZachGawlik, line 48: `<menu>` and `<menuitem>` seem to have [limited support](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem) and simply calling them buttons seemed at least as semantic.

nickoneill: Sounds good to me :)

nickoneill:
This looks great @ZachGawlik! Thanks

nickoneill (approved):

PR #17, opened 2017-01-26
Reduce deployed app.js from 701KB to 107KB

ZachGawlik:
Created `build-scripts` gulp task for deploying a smaller `app.js` by removing browserify debugging and uglifying the result. Locally it chopped off 594KB on the file, so hopefully we'll see some decreased initial page load times once deployed.


nickoneill:
Whoa, that's fantastic. Does this maintain the shims that make things like function arrows `=>` behave in older browsers?

PR #20, opened 2017-01-26
Scroll content into view when activating an issue

rosemarydotworld:
@jameshome I think this is what you were trying to do. This was quick and dirty, so keep an eye out and adjust if anything else needs doing.


jameshome:
Yes, that's perfect. Thank you!

PR #21, opened 2017-01-26
Mobile phone-number formatting

thejbru:
Forgive me, haven't been able to get a local version running so I'm not sure, but I *think* this suggestion makes the phone number to call appear on a single line on phones, adding some padding to give it some breathing room and to make it easier to tap when that gets implemented.


thejbru:
I see this overlaps with the other phone formatting request, so I'll close it.

PR #25, opened 2017-01-26
Refine mobile view

seandellis:
Thanks for building this @nickoneill! 

I've refined the mobile layout with better font sizes on the call page. I've also added a mixin that let's you specify the breakpoint with less code.

The phone size doesn't fill column width, but it's much better than where we started.

Here are some after/before examples:

5calls-example0
5calls-example1


nickoneill:
This looks fantastic @seandellis, thank you! I'll check this out tonight.

PR #32, opened 2017-01-27
Convert string to number for ".toLocaleString()" to work.

beausmith:
Unclear what the replace was replacing. Previous commit didn’t mention why.

## Before

image

## After

image


nickoneill:
THAT'S why toLocaleString wasn't doing what I thought it should

PR #34, opened 2017-01-27
Brevity adds clarity.

beausmith:
## Before

image

## After

image

PR #45, opened 2017-01-28
comprehensive style and copy changes

jameshome:
A new responsive design and corresponding CSS. Also, some minor text improvements.


nickoneill:
First: really great structure and transition between desktop and mobile views. 

- full-width issues on mobile look great!
- adjustments to the location picker are solid, reducing the text in the location area is the right idea, let's do it 100% as @beausmith notes in #34 
- the circular contact photos and text changes nearby look really good, they do a great job at highlighting the name and action (phone number)

I'm 100% for making the text more readable by increasing contrast but I think the colors have gone a bit too far in that direction.

I think we can still accomplish this with a white-on-blue scheme like we currently have.

Colors, logos:

- Let's keep the logo from @stewartsc, that's one of my favorite things from the current design 
- There's something about the star-phone the I don't like... but more so I don't think there's a reason to have a header in the current incarnation, just taking up room that could be content
- Full opacity gray/white sidebar with shadows is MUCH nicer than current design
- Let's try the darker blue background again with light text
	- As discussed on slack, can we remove Obama/Biden or darken the background even more to give more contrast for the text?
- Softer rounded buttons with actual hover states are totally awesome
- I don't think we need the numbered items on the issues, we do organize them by what we consider most relevant at the top but I don't want to assign a numerical importance
- We pick the order and total calls available in airtable, no need to limit them to 5 here (I do agree we need to keep the list relatively short though)

Text:

- The # of calls on the issue list helps people decide that they can do "just 2 more calls", let's bring it back
- Ditto on the "x calls left" on the bottom of each contact
- Good call on removing the double "why calling works" and "about us" links, I've heard a few complaints about that
- I do like the total calls count repeated on the #done page, was there a specific reason to remove it?
- What about "together we've made x calls" rather than "x calls so far" or "x call to date"?

Code:
- Thank you for unifying the promote box :)
- The css improvements are amazing and now I no longer have to pretend I know what a rem is

In general, awesome set of changes that I'm sure we can make happen. My biggest concern was that it felt like a "redesign" and it should be a refinement of what we have and is working well. Upon looking at the code it's clear that it really is a refinement on what we have, the visual style just needs to reflect that so we don't scare away people who have made their calls and plan on coming back.

PR #61, opened 2017-01-30
Add governor type contact

### LINKED ISSUE #28: Pull Governor information from the civic API
Currently we pull local rep information from the google civic api and return them to users when airtable issues have the US SENATE or US HOUSE contacts.

Similarly, calling the state governors office is useful in certain issues such as rejecting demands that states shut down sanctuary cities, etc. We should do a similar thing to pull governors from the civic API (in the same call as local reps, if possible) and replace when airtable has a GOVERNOR contact.

agjohnson:
This adds a governor type contact to the contact list returned by issues API. If
contact name == `GOVERNOR`, this lookup will take place.

And, uh, pardon my Go etiquette, this is all new to me.

Fixes #28

### DIFF
--- a/go/civic.go
+++ b/go/civic.go
@@ -15,15 +15,19 @@ import (
 )
 
 const (
-	roleLowerBody = "legislatorLowerBody"
-	roleUpperBody = "legislatorUpperBody"
-	roleLevel     = "country"
-	areaHouse     = "House"
-	areaSenate    = "Senate"
+	roleLowerBody        = "legislatorLowerBody"
+	roleUpperBody        = "legislatorUpperBody"
+	roleHeadOfGovernment = "headOfGovernment"
+	roleLevelCountry     = "country"
+	roleLevelState       = "administrativeArea1"
+	areaHouse            = "House"
+	areaSenate           = "Senate"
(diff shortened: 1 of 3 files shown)


go/civic.go
```diff
@@ -15,15 +15,19 @@ import (
...
-	areaHouse     = "House"
-	areaSenate    = "Senate"
+	roleLowerBody        = "legislatorLowerBody"
+	roleUpperBody        = "legislatorUpperBody"
+	roleHeadOfGovernment = "headOfGovernment"
+	roleLevelCountry     = "country"
+	roleLevelState       = "administrativeArea1"
+	areaHouse            = "House"
+	areaSenate           = "Senate"
+	areaGovernor         = "Governor"
```
agjohnson, line 25: Not sure if `area` is used for anything else, but `area` == `Governor` doesn't make much sense at all. Anything a better fit here?

nickoneill: `area` is as intended here, just a general placeholder for where representatives serve

nickoneill:
This is great! Planning on using this early next week for an issue to Governors!

screen shot 2017-02-02 at 10 38 21 pm

PR #79, opened 2017-01-31
Scroll to address input

rboyce:
On mobile browsers, tapping "set your location" at the bottom of the page has no visible effect since the address field is scrolled offscreen. This diff scrolls the address input into view when "set your location" is tapped. This requires the form element to be in the DOM, so it is hidden when the user doesn't need to see it. Verified this behavior in desktop Chrome as well, though scrolling is only necessary on small screens.

iOS simulator screenshot

### DIFF
--- a/static/js/components/issuesLocation.js
+++ b/static/js/components/issuesLocation.js
@@ -4,12 +4,13 @@ module.exports = (state, prev, send) => {
   return html`
     <div class="issues__location">
     ${pretext(state)}
+    ${addressForm(state)}
     </div>
   `;
 
   function pretext(state) {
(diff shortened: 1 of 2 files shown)


nickoneill:
Thanks @rboyce! We've got more than a few complaints about this, I will take a look today

nickoneill:
Looks good! You caught us in a weird spot last night, I ended up reverting a few commits related to geolocation. Can you rebase and we'll bring this in?

rboyce:
@nickoneill Rebased!

nickoneill:
Thanks!

PR #99, opened 2017-02-02
Browser Location (version 2)

capndesign:
This fixes the looping issues and is properly merged with master.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -107,8 +107,8 @@ app.model({
       return { totalCalls: totals.count }
     },
     receiveIPInfoLoc: (state, data) => {
-      geo = response.loc
-      city = response.city
+      geo = data.loc
+      city = data.city
       time = new Date().valueOf()
       store.replace("org.5calls.geolocation", 0, geo, () => {});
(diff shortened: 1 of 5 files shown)


static/js/main.js
```diff
@@ -87,25 +106,14 @@ app.model({
...
-          return { geolocation: geo, cachedCity: city, geoCacheTime: time }
-        } else {
-          Raven.captureMessage("Location with no city: "+response.loc, { level: 'warning' });
-        }
-      } catch(e) {
-        Raven.setExtraContext({ json: data })
-        Raven.captureMessage("Couldn't parse ipinfo json", { level: 'error' });
-      }
+    receiveIPInfoLoc: (state, data) => {
+      geo = response.loc
```
nickoneill: Where is `response` coming from? It works but I do not understand why!

nickoneill:
Just that one comment that I want to understand, looks awesome and handles lots of different weird cases!

capndesign:
@nickoneill I think this was a magical fat arrow that saved us! I changed `response` to `data`, so we should be good.

nickoneill:
Awesome change! This looks great on production

PR #101, opened 2017-02-02
Informs the user they've entered an invalid address/zip

### LINKED ISSUE #88: Address/zip error state if the location isn't valid
What information can we pull back form google civic api to determine that the location isn't valid?

jadametz:
The Civic API returns `400` [responses](https://developers.google.com/civic-information/docs/v2/standard_errors) for many reasons. Of those is a `parseError` where they return "Failed to parse address" for invalid addresses.

In the event that the user enters a zip or address that the Civic API was unable to parse, we present that information to the user and ask them to try again.

This should resolve #88

### DIFF
--- a/go/handler.go
+++ b/go/handler.go
@@ -8,6 +8,8 @@ import (
 )
 
 const (
+	addrFailedToParse = "400 Failed to parse address"
+	addrNoInfo        = "404 No information for this address"
 	localPlaceholder  = "LOCAL REP"
 	senatePlaceholder = "US SENATE"
 	housePlaceholder  = "US HOUSE"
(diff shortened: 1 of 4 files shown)


nickoneill:
This looks good @jadametz! Can we also clear the previous entry in the location text box when it's re-shown? Helps reinforce the idea that the old location was invalid.

nickoneill:
Two things here:

* `queryEscape` on line 221 of `civic.go` double-encodes the address so things like lat/lng return a failed address, just removing that escape should be fine
* I tried merging this on my local machine and it looks like it conflicts with the recent work from @josharian that is in master - take a quick shot at rebasing?

Sorry, second time run-around. Sometimes that's how it works in open source 🙃

nickoneill:
Fixed the queryescape thing in another commit... I don't understand why github doesn't show a merge conflict here? I'll check on it again tomorrow.

jadametz:
Do you have the latest branches of everything on your local machine? My last commit merged `master` into my branch to fix what conflicts did exist in `civic.go`. 

Because these branches have been made public, I don't think a rebase is what you want to do, is it?

nickoneill:
Nope, it was just my local problem. Good to go now!

PR #132, opened 2017-02-03
Add field office phone information to backend response

josharian:
The first commits in this series are preliminary cleanup.

The last commit adds the meat. See its commit message for details.

Still to do:

* Hook up the UI to use the new response information. Help with that would be great. [Sample response gist](https://gist.github.com/josharian/714ab138ed875a5742e86ea30e2303a4) and see the changes to model.go in the last commit for docs about the fields in the response.
* Record which phone number people called when we register their call and the result.
* Hook up the UI to send the phone number people called. This can happen even before the backend is ready to receive it.


go/field_phones.go
```diff
@@ -0,0 +1,751 @@
...
+		"32399": "8505999100", // Tallahassee
+		"33609": "8132875035", // Tampa
+		"33418": "5617753360", // Palm Beach Gardens
+		"33166": "3054188553", // Doral
+		"32502": "8504332603", // Pensacola
+	},
+	// Johnny Isakson (R-GA)
+	"2022243643": map[string]string{
+		"30339": "7706610999", // Atlanta
+		"":      "4042911128", // Atlanta
```
beausmith: missing key?

go/field_phones.go
```diff
@@ -0,0 +1,751 @@
...
+	},
+	// Elizabeth Warren (D-MA)
+	"2022244543": map[string]string{
+		"02203":  "6175653170", // Boston
+		"01103 ": "4137882690", // Springfield
+	},
+	// Benjamin L. Cardin (D-MD)
+	"2022244524": map[string]string{
+		"21201":     "4109624436", // Baltimore
+		"    20720": "3018600414", // Bowie
```
beausmith: extra spaces?

go/field_phones.go
```diff
@@ -0,0 +1,751 @@
...
+		"21502 ":    "3017772957", // Cumberland
```
beausmith: extra space?

josharian:
I don't know how important it is that people call their closest field office vs any of them.

@rebeccakaufman do you know? If it is important, I can grab some zip code lat/longs and try harder to pick out the closest one.

beausmith (commented):
Nice work. I found a couple data scrubbing issues.

beausmith:
@josharian: @rebeccakaufman and are are in my kitchen chatting through ideas. Some thoughts:

- According to @rebecca DC office typically handles policy, while local offices handle other more local issues. Thus, the DC number should always be primary. She has heard this from a friend who is a on the staff with a senator.

- DC is currently getting a very high volume of calls, so providing alternate local office numbers is useful right now. Quickest way to do this is to just have some toggle to reveal these additional numbers under the main number. 

_But what about recording which field office a user called?_

If the DC call volume doesn't go down and calling field offices is very relevant and helpful, then let's add tracking to include phone number. At present, its more important to just get the alternate phone numbers on the page.

josharian:
Fixed the data quality issues. Thanks.

> the DC number should always be primary

Yep. That was the idea behind separating `hq` in the response from the field offices. The UI can show `hq` first, and then if they report that the office was unavailable, the UI can switch to one of the other phone numbers. All the data should be there in the response.

> Quickest way to do this is to just have some toggle to reveal these additional numbers under the main number.

That works too. My goal here is to make the data available to the front-end and let you decide how to present. :)

> let's add tracking to include phone number

Working on that right now. :)

beausmith:
Love it.

beausmith:
@josharian - Can we put the city and state into the static phone data support something like this.

Clicking on "View all numbers for Dianne Feinstein" would toggle the list of numbers below.

(Obviously wrong data for Feinstein 😉 )

---

image

---

Obviously, we can do a simple solution without city and state first.

josharian:
Yep. On it.

josharian:
OK, PTAL.

Once I have the reporting hooks in place, I will update the data file again with the new source that @rebeccakaufman just sent me.

josharian:
Reporting hooks also now pushed up and ready for feedback. Now looking at improving the list of field offices.

josharian:
Updated to better data from @rebeccakaufman. As far as I know right now, this is ready for review + deployment if it looks good. Front-end use of this data can happen and go out independently.

Please take another look, @beausmith. Thanks for all the feedback so far.

I can also squash this PR if you'd like.

beausmith:
Cool. I'll check it out later tonight, but no need to wait for my blessing. 😁

go/civic.go
```diff
@@ -169,6 +183,31 @@ func (x *Official) ID() string {
 	return spaceReplacer.Replace(x.Name)
 }
 
+func mapDigitsOnly(r rune) rune {
+	if !unicode.IsDigit(r) {
```
bradfitz: overkill.

   return '0' <= r && r <= '9'

Unless you really want to deal with every character in  http://www.fileformat.info/info/unicode/category/Nd/list.htm everywhere else and be paranoid about them.  I think you'd prefer to exclude them.

Also, I'd just write this func inline in digitsOnly.

go/civic.go
```diff
@@ -149,18 +151,30 @@ func (x *Office) Area() string {
...
+	for z, fieldPhone := range senateFieldPhones[clean] {
+		formatted := formatphone(fieldPhone)
+		if z == zip {
+			p.BestField = formatted
+			continue
+		}
+		p.AllField = append(p.AllField, formatted)
 	}
-	return reformattedPhone(x.Phones[0])
+	return &p
```
bradfitz: We normally dissuade people from using this style, because it leaves `p` vulnerable to being accidentally copied between the point of declaration and here. 

Preferred is just `p := new(Phones)` above

go/models.go
```diff
@@ -19,6 +19,13 @@ func (i *Issue) String() string {
 	return asJson(i)
 }
 
+// Phones are ways to call
```
bradfitz: All three of these fields are unclear about what their actual values look like.

Are these strings like "Washington DC office", phone numbers like "503-123-4567", or phone numbers like "12065551212" or?

go/civic.go
josharian: Done.

go/civic.go
josharian: Good to know. No longer relevant for this PR, due to subsequent commits. (I miss gerrit already.)

go/models.go
josharian: Added some docs.

nickoneill:
top notch PR A+++ 💯💯💯 would merge again

Issue #131, opened 2017-02-03
Add to JavaScript unit test coverage


cdoremus:
I'm creating a number of unit tests for components including call, contact, done, issuesList, issuesListItem and script. They follow the pattern I used with callcount_test.js in PR #97.

cdoremus:
Since the web site content is in flux, I am trying to not go overboard with test coverage, but making sure most of the bases are covered. Another complication is the nodes returned by choo's html tagged templates does not appear to implement a lot of the DOM API, and I thought it would be overkill at this point to use Karma or Mochify to run the tests in a browser. At any rate, FWIW, I feel they are a good starting point and will be useful when CI is added to the project.

cdoremus:
Rob (@Mr0grog) as you suggested in #133, put together a PR configuring an in-browser testing solution of your choosing that runs off the Gulp test task and works with the callcount and localstorage tests you have in the branches you referenced in #133. When that is done, I'll migrate the tests in #134. Thanks, in advance.

Mr0grog:
@cdoremus I was unable to get to this today, but will see what I can get done tomorrow evening (2/6).

Mr0grog:
@cdoremus Sorry I’ve been so slow to get around to finishing this. I hope I haven’t held you up in any way. Preliminary PR is over at #181.

Mr0grog:
Should this be closed, since #134 and #181 are done and merged? Any additional work this is meant to cover?

cdoremus:
As @Mr0grog stated, work done on #134 and #181 implements this issue.

PR #133, opened 2017-02-03
Suppress quota exceptions in localStorage

### LINKED ISSUE #126: Site is unusable in Safari’s private browsing mode
I keep my phone (iOS) in private browsing mode, but can’t use the site at all because the JS fails to handle exceptions from local storage. (This also affects desktop, but private browsing isn’t my default there.)

Specifically, Safari will throw a `QuotaExceededError` when setting a value to local storage in private browsing mode. (https://www.w3.org/TR/webstorage/#dom-storage-setitem) This happens right away when the page is loaded and geolocation permissions are either requested or denied. The rest of the application code stops loading and a user is then pretty much stuck :\

I’m happy to work on a quick PR for this.

Mr0grog:
This fixes #126 by suppressing any exceptions when setting items in `localStorage`. (In private browsing mode, Safari sets the storage quota to 0, which means localStorage throws an exception whenever you try to set anything.)

I started to write tests as well, but realized that required changing the test process significantly: it needs to run in a browser or needs a tool like [`mocha-jsdom`](https://www.npmjs.com/package/mocha-jsdom) or [`jsdom-global`](http://npmjs.com/package/jsdom-global), though the latter seems to conflict with other tests. I’m happy to do that work, too, if you want it.

### DIFF
--- a/static/js/utils/localstorage.js
+++ b/static/js/utils/localstorage.js
@@ -10,21 +10,30 @@ module.exports = {
   add: (storeName, item, cb) => {
     module.exports.getAll(storeName, (items) => {
       items.push(item)
-      window.localStorage[storeName] = JSON.stringify(items)
-      cb()
+      saveStore(storeName, JSON.stringify(items), cb)
     })
   },
   replace: (storeName, index, item, cb) => {
     module.exports.getAll(storeName, (items) => {
       items[index] = item
-      window.localStorage[storeName] = JSON.stringify(items)
-      cb()
+      saveStore(storeName, JSON.stringify(items), cb)
     })
   },
   remove: (storeName, cb) => {
     module.exports.getAll(storeName, (items) => {
-      window.localStorage[storeName] = null
+      window.localStorage.removeItem(storeName)
       cb()
     })
   },
+}
+
+// handle storage quota errors
+function saveStore (storeName, storeItems, cb) {
+  try {
+    window.localStorage[storeName] = storeItems
+    cb()
+  }
+  catch (e) {
+    cb(e)
+  }
 }


nickoneill:
Thanks! I was wondering where these exceptions come from, I am not super familiar with localstorage. I'll check this out in detail tonight.

Mr0grog:
No problem; happy to help out :)

nickoneill:
I have no strong opinions on how to do js testing and I also do not know the best approaches, if you want to open a separate issue with this and tag @cdoremus who has written a few basic tests for us, I would love to have more tests in the codebase.

cdoremus:
@Mr0grog (Rob), my first motivation was to get code coverage above zero in this project the easiest way possible since the app is evolving rapidly with much hard coded UI content, so tests are likely to be unstable. Given that Choo uses pure functions to produce DOM-like nodes, I decided to do unit testing without the browser. I figured that we would add in-browser integration tests later for things like local storage and e2e tests in various web clients. I have worked with Karma/Jasmine/Mocha to test Angular and Jest to test React code, but I have no experience with Choo testing and there are practically no examples out there. If you can point me to a forked branch with your test code and test configs or attach it to this or another issue, I can work with you on converting my tests to in-browser tests. Thanks, in advance.

Mr0grog:
@cdoremus I am totally down for helping to improve tests. Here are some quick notes, but let's start a separate issue for this.

> I have worked with Karma/Jasmine/Mocha to test Angular and Jest to test React code, but I have no experience with Choo testing and there are practically no examples out there.

Heh, I had looked briefly at Choo before, but have no real experience with it, so I’m in a pretty similar boat here ;)

> If you can point me to a forked branch with your test code

I had tried to keep it pretty minimal, so didn’t go as far as in-browser testing. I did two simple approaches using various flavors of jsdom (a full HTML5 DOM parser in JS—it creates a browser-like environment in Node, though it is still short of the real thing):

- [`localstorage-tests-with-mocha-jsdom`](https://github.com/5calls/5calls/compare/master...Mr0grog:localstorage-tests-with-mocha-jsdom) Uses `mocha-jsdom` to run the localstorage tests with JSDom. It keeps the DOM stuff contained to the tests that use it, though that adds some rigamarole to setting up each test.

- [`localstorage-tests-with-jsdom-global`](https://github.com/5calls/5calls/compare/master...Mr0grog:localstorage-tests-with-jsdom-global) uses `jsdom-global` and automatically installs all the jsdom globals when Mocha boots so all code sees normal browser globals even though it’s in Node. That required some changes to the tests for Choo components, though, since you get different (more browser-like) results when running on top of JSDOM.

If you want to have better in-browser testing, though, I think we should not use either of those solutions above—we should actually get this running in a browser, whether that’s a simple as PhantomJS/SlimerJS/NightmareJS directly or more complicated like using Webdriver or Karma.

Should we create a new issue for this or do it as part of #131?

Mr0grog:
(Also, we should see about setting up CI—I have experience with both Travis and Circle; I like Circle much more.)

cdoremus:
Thanks, Rob. It looks like ```jsdom-global``` is the best option. Do you agree? I should have some more time to look at this tomorrow evening.

Mr0grog:
If we’re trying to keep it quick and simple, I think `mocha-phantomjs` or `gulp-mocha-phantomjs` would be better—it’s not much more complicated and gets the front end tests running in an actual browser engine. I can throw together a simple PR for that.

Ultimately, though, I think Sauce/Browserstack/Karma/Webdriver is the best thing—browser code should be tested in the actual browsers it will run in—but setting that up is more complicated and has more room for debate.

nickoneill:
Can you move this convo to #131 so it has better visibility?

cdoremus:
Rob (@Mr0grog) -- as Nick (@nickoneill) suggested -- I have continued this discussion in #131.

PR #137, opened 2017-02-04
Display field office numbers (hidden behind toggle)

beausmith:
More numbers to call!!

field-office-labels

### DIFF
--- a/static/js/components/contact.js
+++ b/static/js/components/contact.js
@@ -10,7 +10,26 @@ module.exports = (c, state, prev, send) => {
     repID = c.party.substring(0,1) + "-" + c.state;
   }
 
-	return html`
+  let fieldOffices
+  if (!!~c.field_offices) {
+    fieldOffices = html`
+      <p class="call__contact__show-field-offices"><a onclick=${() => {send('toggleFieldOfficeNumbers')}}>Call a local office</a></p>
+    `
+    if (state.showFieldOfficeNumbers) {
+      fieldOffices = html`
+        <div>
+          <h3 class="call__contact__field-offices__header">Field offices for ${c.name}:</h3>
+          <ul class="call__contact__field-office-list">
+            ${c.field_offices.map(office => html`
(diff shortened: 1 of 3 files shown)


static/js/components/contact.js
```diff
@@ -10,14 +10,34 @@ module.exports = (c, state, prev, send) => {
     repID = c.party.substring(0,1) + "-" + c.state;
   }
 
-	return html`
+  let fieldOffices
+  if (!!c.field_offices) {
```
josharian: Stupid question: Does `!!` treat empty array as identical to missing array? (It should, just to be safe.)

static/js/components/contact.js
```diff
@@ -10,14 +10,34 @@ module.exports = (c, state, prev, send) => {
...
+    fieldOffices = html`
+      <p class="call__contact__show-field-offices">Busy line? <a onclick=${() => {send('toggleFieldOfficeNumbers')}}>View field office numbers for ${c.name}</a></p>
```
josharian: Maybe just "Call a field office."? I'm happy to defer to others on the wording.

josharian (commented):
Thanks for whipping this up. Two small questions. @nickoneill , this one is yours now. :)

static/js/components/contact.js
nickoneill: Yes, I think the duplication of the name here is unnecessary. Also, I'm more partial to "local office" terminology

static/js/components/contact.js
beausmith: Gonna update to use the 'bitwise' operator so that…

image

Any use case I'm missing? @josharian

static/js/components/contact.js
beausmith: Ok… will update to "Call a local office"

static/js/components/contact.js
josharian: Looks right

beausmith:
@nickoneill Updated, rebased, and ready to merge.

nickoneill:
👏

PR #139, opened 2017-02-04
Ignore custom routing for privacy link.

### LINKED ISSUE #113: Privacy Link
On chrome Version 55.0.2883.87 m (64-bit) on Windows 10, when I click the privacy link, the page content doesn't change.  The address bar shows the correct link to https://5calls.org/privacy.html, but for some reason no forwarding is happening.  Other two links in footer work as expected.  Odd issue, but I thought I would share it.

beausmith:
https://github.com/yoshuawuyts/sheet-router#href

I think this will work… but since it's broke on production already… ship it!

Resolves #113

### DIFF
--- a/static/html/index.html
+++ b/static/html/index.html
@@ -100,7 +100,7 @@
         <a href="https://twitter.com/make5calls">
           <i class="fa fa-twitter" aria-hidden="true"></i> Updates on Twitter
         </a>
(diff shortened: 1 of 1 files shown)


nickoneill:
Thanks for... actually looking at the docs! I didn't quite get that far last night.

Issue #24, opened 2017-01-26
Explain 1 issue per call in FAQ


Ipstenu:
It seems ineffective to call the same person X times about issues. It will fill up mailboxes faster, which is already a problem. Perhaps allowing to sort by contact person could help this.

(PS, thank you for this work! The scripts help anyone with anxiety know what to do!)

nickoneill:
Thanks @Ipstenu! It's more effective right now (according to staffers) to provide a single issue per phone call / voicemail. Hopefully the local offices support in #12 will alleviate this for the moment.

Ipstenu:
That makes sense and sounds like a FAQ page may need this information :)

nickoneill:
You're totally right. Let's keep this as a reminder.

oderby:
Please include a reference/source in your explanation. It goes against my intuition and much of the (equally source-less) advice I currently see swirling around in my social circles (so as a user, I'm skeptical right now of using the site as is).

nickoneill:
Hi @oderby, we're in contact with senate staffers who are actually in charge of answering the phone so that's where we're getting this info right now. Will update you if we hear that anything changes.

nickoneill:
This is covered in the FAQ

PR #146, opened 2017-02-05
About Page: adds explanation for organizing calls by issue instead of representative

lauren:
In response to issue #24


nickoneill:
We want to add this info on a new scripts page, I'll hold off until that comes around.

PR #148, opened 2017-02-05
Makes the field offices link more actionable:

lauren:
* increases font size
* adds pointer when hovering over link
* updates copy to explain why you might want to call a field office


nickoneill:
Hey @lauren, I just pulled in something similar that was previously in review. Send you an invite to the slack if you want to coordinate more closely

nickoneill:
Made this more prominent independently - thanks for the suggestion!

Issue #85, opened 2017-01-31
Feature Request: dynamic call scripts.


issackelly:
It would be really nice to have dynamic call scripts so I have to make fewer intepretations while I'm on the phone.

Good!

```
Hi, my name is [Name] and I'm a constituent from [City, State, Zip].

I’m calling to tell [Senator/Rep’s Name] that our community does not support Trump’s Muslim ban, and we expect [him/her] to take an immediate stand against this unjust executive order.

[Optional: add anything else you’d like here]

Thank you for your time.

[IF LEAVING A VOICEMAIL: please leave your entire home address to ensure your call is tallied]
```

Better?

```
Hi, my name is [Name] and I'm a constituent from [City, State, Zip].

I’m calling to tell Senator Feinstein that our community does not support Trump’s Muslim ban, and we expect her to take an immediate stand against this unjust executive order.

[Optional: add anything else you’d like here]

Thank you for your time.

[IF LEAVING A VOICEMAIL: please leave your entire home address to ensure your call is tallied]
```

Better still?

```
Hi, my name is Issac and I'm a constituent from Oakland, CA 94618

I’m calling to tell Senator Feinstein that our community does not support Trump’s Muslim ban, and we expect her to take an immediate stand against this unjust executive order.

[Optional: add anything else you’d like here, (ask about second issue)]

Thank you for your time.

[IF LEAVING A VOICEMAIL: please leave your entire home address to ensure your call is tallied]
```

issackelly:
I could see storing my name, address, and zipcode in localStorage or a cookie. 
I could see the script area being an text box or ace editor so that I can make my adjustments before the phone call is made.

Being on the phone isn't always comfortable! Having a script is nice. You're doing great work.

nickoneill:
Mentioned on slack and I'll reiterate here: this is great as long as we do it automatically based on the info supplied. Should be able to get a bunch of this via the civic api `normalizedInput` field, let me know if you need more in the API response.

joshfeingold:
I think with this one it should be dependant.  If the app has the info, like the senator name, that def makes sense, but the individuals name, I think requires that the user add their info which is another step in the process to get them actually making an impact.  The less steps that better.

jameshome:
See #19 for some related ideas, mostly covered here (closing that one out).

samhavens:
Gender is a hiccup here. The scripts sometimes contain `[him/her]`, but that information doesn't seem to be available from the Civics API. For example, this shows up in this script:
```
[...]

PR #150, opened 2017-02-05
Dynamic call scripts part 1 - fill in representative info

samhavens:
This is in response to #85, though it may not cover all of the desired behavior.

*Summary:*

- Add dependency on `gender-detection` package (see discussion below)
- Change signature of `script` component to take `currentContact`
- Add templating to script component to fill in the user's current contact's information

*Example:*

Prior behavior:

> Hi, my name is [Name] and I'm a constituent from [City, State, Zip].
>
>I’m calling to tell [Senator/Rep’s Name] that our community does not support Trump’s Muslim ban, and we expect [him/her] to take an immediate stand against this unjust executive order.
>
>[Optional: add anything else you’d like here]
>
>Thank you for your time.

With this change:

> Hi, my name is [Name] and I'm a constituent from [City, State, Zip].
>
>I’m calling to tell Ms. Feinstein that our community does not support Trump’s Muslim ban, and we expect her to take an immediate stand against this unjust executive order.
>
>[Optional: add anything else you’d like here]
>
>Thank you for your time.

*Discussion:*

I was really reluctant to infer gender. I have seen some fights in other open-source projects about this exact issue. However, without doing so, the filled-in scripts just felt awkward. I feel like the point of this modification is to make the process even _easier_ for people who are uncomfortable on the phone, and giving them an awkward-sounding script wouldn't accomplish that. Also, currently, all representatives are gender binary, so that's a factor too.

So, why this particular package for inferring gender? It doesn't use `fs`, so it can be browserified. There may be other options, I examined a few, and this one seemed ok.

*Todo:*

Still need to store the user's information in localStorage and fill that in as well.

### DIFF
--- a/static/js/components/script.js
+++ b/static/js/components/script.js
@@ -1,16 +1,31 @@
 const html = require('choo/html');
 const find = require('lodash/find');
+const gender = require('gender-detection');
 const scriptLine = require('./scriptLine.js');
 
 module.exports = (contact, state, prev, send) => {
 	  const issue = find(state.issues, ['id', state.location.params.issueid]);
 	  const currentContact = issue.contacts[state.contactIndex];
 
+    const repGender = gender.detect(contact.name);
+    const titleMap = { male: 'Mr. ', female: 'Ms. ', unisex: '' };
+    const pronounMap = { male: 'him', female: 'her', unisex: 'them' };
+    const repTitle = titleMap[repGender];
+    const formalName = `${repTitle}${contact.name.split(' ').pop()}`;
+
+    const scriptFormatter = (issue) =>
+      issue.script.replace('[him/her]', pronounMap[repGender])
+                  .replace('[Senator/Rep’s Name]', formalName)
+                  .replace('[Senator/Rep\'s Name]', formalName)
+                  .replace('[Senator’s Name]', formalName)
+                  .replace('[Senator\'s Name]', formalName)
+                  .split('\n').map((line) => scriptLine(line, state, prev, send));
+
     if (currentContact != null) {
       return html`
       <div class="call__script">
         <h3 class="call__script__header">Your script:</h3>
-        <div class="call__script__body">${issue.script.split('\n').map((line) => scriptLine(line, state, prev, send))}</div>
+        <div class="call__script__body">${scriptFormatter(issue)}</div>
       </div>`
     } else {
       return html``
--- a/static/js/components/script.js
+++ b/static/js/components/script.js
@@ -2,16 +2,16 @@ const html = require('choo/html');
 const find = require('lodash/find');
 const scriptLine = require('./scriptLine.js');
 
-module.exports = (state, prev, send) => {
+module.exports = (contact, state, prev, send) => {
 	  const issue = find(state.issues, ['id', state.location.params.issueid]);
 	  const currentContact = issue.contacts[state.contactIndex];
-    
+
     if (currentContact != null) {
       return html`
       <div class="call__script">
         <h3 class="call__script__header">Your script:</h3>
         <div class="call__script__body">${issue.script.split('\n').map((line) => scriptLine(line, state, prev, send))}</div>
-      </div>`      
+      </div>`
     } else {
       return html``
     }
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   "dependencies": {
     "choo": "^4.1.0",
     "connect-logger": "0.0.1",
+    "gender-detection": "0.0.5",
     "gulp-util": "^3.0.8",
     "http-server": "^0.9.0",
     "lodash": "^4.17.2",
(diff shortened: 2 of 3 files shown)


nickoneill:
Thanks @samhavens! We'll communicate to the writing team to standardize on these replacement strings. 

In the meantime, something like "lindsey graham" is incorrectly identified. I expect there are only a handful of these in our dataset but we should have some way of overriding them if a user tells us they're incorrect (or if they pop to the top of the head, like Senator Graham).

samhavens:
Good point. I was thinking, it may be easier to distinguish Senator vs
Representative?

nickoneill:
Yeah, that might make things easier @samhavens. There should be an `area` field for each contact noting if they're in the House, Senate or Other.

static/js/components/script.js
```diff
@@ -1,18 +1,33 @@
...
-    
+
+    const repGender = gender.detect(contact.name);
+    const titleMap = { male: 'Mr. ', female: 'Ms. ', unisex: '' };
+    const pronounMap = { male: 'him', female: 'her', unisex: 'them' };
+    const repTitle = titleMap[repGender];
+    const formalName = `${repTitle}${contact.name.split(' ').pop()}`;
+
+    const scriptFormatter = (issue) =>
+      issue.script.replace('[him/her]', pronounMap[repGender])
```
oderby, line 17: It seems it's not currently used, but why not also support [he/she] substitution?

static/js/components/script.js
```diff
@@ -1,18 +1,33 @@
...
 
-module.exports = (state, prev, send) => {
+module.exports = (contact, state, prev, send) => {
 	  const issue = find(state.issues, ['id', state.location.params.issueid]);
 	  const currentContact = issue.contacts[state.contactIndex];
-    
+
+    const repGender = gender.detect(contact.name);
+    const titleMap = { male: 'Mr. ', female: 'Ms. ', unisex: '' };
+    const pronounMap = { male: 'him', female: 'her', unisex: 'them' };
```
oderby, line 12: style nit/question: Do we care about putting LUT like this outside the function (to line 5)?

static/js/components/script.js
```diff
@@ -1,18 +1,33 @@
 const html = require('choo/html');
 const find = require('lodash/find');
+const gender = require('gender-detection');
 const scriptLine = require('./scriptLine.js');
 
-module.exports = (state, prev, send) => {
+module.exports = (contact, state, prev, send) => {
 	  const issue = find(state.issues, ['id', state.location.params.issueid]);
 	  const currentContact = issue.contacts[state.contactIndex];
```
oderby, line 8: seems like `contact === currentContact` (line 8 here is same as line 26 in call.js). Either delete line 8 or don't need to add the argument.

Oh, just noticed line 7 here is same as line 11 above...

And looks like we're passing contact by argument to contact.js in 30 in call.js...

Seems like we should decide how we want to handle these redundant state lookup/extraction needs. It seems we currently have two different practices: look it up once, and pass by argument downwards (the case of `contact` here and contact.js) vs. look it up independently in every component that needs it (e.g. `issue` here and call.js). My intuition is to require all components to have a standardized interface (that being `(state, prev, send)`) and providing helpers/utilities for looking up specific/common states from the state object if that becomes onerous. But, I'm new to Choo/this code base, so will happily defer to @nickoneill or others on which we should follow. But whatever we choose, we should document and be consistent.

oderby (commented):
Do we care about using the appropriate title for married women? (Ms. vs Mrs.)? Should we just avoid it by using their other title of 'Senator' or 'Representative'? I guess 'Representative' is long... but could just shorten to 'Rep.'. There's also consideration for other people outside of Senators and Representatives (President, Judge, Governor)...if we have any ambitions of calling those folks, we might want to consider a more robust treatment of titles (and how they're indicated in scripts)

nickoneill:
@samhavens let me know if you want to make the proposed changes, otherwise I'll hand it off to someone else.

samhavens:
I am happy to. Sorry, I am new to the project and don't know how long to wait for consensus.

Want me to encapsulate the gender info more, add he/she map, but change the actual script replacements to be to Rep. and Senator? Also, what's the decision on changing the signature of the exported function?

static/js/components/script.js
nickoneill: Yep, I think this is just a case of me trying it different ways. Let's standardize on the lookup and keep the signature the same unless it makes things significantly more difficult for some case.

nickoneill:
Yep - those sound like reasonable changes (ignoring the title issue mentioned by @oderby - let's do Senator / Rep. [Last Name] for names and then just he/she if necessary). Signature talk inline 👆

nickoneill:
🏓

bernardwang:
Hi @samhavens, any updates on this PR? If you don't want to make these changes I would be happy take the issue!

samhavens:
@bernardwang I totally dropped the ball on this and would love your help. Thank you.

nickoneill:
Closed in favor of #299

PR #154, opened 2017-02-05
Properly report number of remaining calls

### LINKED ISSUE #78: "You've made all the calls for this issue" listed even when I haven't yet
I can click on an issue for which I have not made calls (no green check), and after the call response buttons I see the text "You've made all the calls for this issue."

Then I tried a hard refresh (ctrl-shift-r) and each issue now says "1 more person to call for this issue", no matter if I've called yet or not.

I'm not sure what is going on, but it seems the message is not always being applied correctly.

This is Chrome on Mac. Let me know if I can provide more info!

phoebejaffe:
Addresses the first UI bug in task #78, but not the refresh-related issue.

The original string "You’ve made all the calls for this issue." doesn't ever appear correctly -- the "Great Work" section handles it nicely without the string, so I removed it.

### DIFF
--- a/static/js/components/outcomes.js
+++ b/static/js/components/outcomes.js
@@ -8,7 +8,7 @@ module.exports = (state, prev, send) => {
   const contactsLeft = issue.contacts.length - (state.contactIndex + 1);
   const callsPluralization = contactsLeft > 1 ? "people" : "person";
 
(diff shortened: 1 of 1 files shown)


beausmith:
LGTM!

Issue #157, opened 2017-02-06
Set debug mode based on localStorage


benjaffe:
This would allow for debug mode to be persistent, and nobody would ever accidentally commit `const debug = true` :)

nickoneill:
No one has ever done a thing like that, least of all me. Of course.

oderby:
Can this be closed now?

PR #158, opened 2017-02-06
Set debug mode based on localStorage

phoebejaffe:
Addresses #157

To turn on debug mode, simply set the localStorage key to true in the console:
```
localStorage['org.5calls.debug'] = 'true'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,13 @@ The other main Gulp task is the `deploy` task, which does not watch for
 changes, and applies additional transforms on the assets -- such as an uglify
 transform on Javascript sources.
 
+To turn on/off debug mode, which adds some reset buttons throughout the interface, run the following in your console:
+
+```
+localStorage['org.5calls.debug'] = 'true' // turn on debug mode
+localStorage['org.5calls.debug'] = 'false' // turn off debug mode
+```
+
 ### Application Server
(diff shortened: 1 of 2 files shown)


phoebejaffe:
Note that the couple of console.logs will disappear unless you turn on "debug mode" with the incantation above :)

beausmith:
@benjaffe - I love this solution. 

I've been meaning to inquire as to if anyone is using these console messages… as I was going to "solve" this by removing these console.log lines.

My thinking is that these are only useful to someone who is debugging the geolocation code… and since that is not an ongoing effort they are no longer useful. If someone is debugging geolocation in the future they can add debugging in… and then remove it before they submit a PR.

Agree/Disagree. Thoughts?

nickoneill:
Yeah, that's the right way to approach this @beausmith. These should never really be in prod in the first place. The debug framework is good though!

phoebejaffe:
I agree that generally logging shouldn't be committed, but if it's innocuous, it's kinda nice to see it doing high-level things when debug mode is on. Makes it easier to trace, possibly generally debug.

So IMO, 'getting geoposition' and 'selecting issue abcdef' are okay, but `console.log('yo1');` isn't :)

Issue #144, opened 2017-02-04
Location Query doesn’t fire on about page


beausmith:
## Repro Steps

1. Clear localStorage in browser
2. Go to <https://5calls.org/#about>

## Result 

You will see this "getting your location"

![image](https://cloud.githubusercontent.com/assets/28444/22621545/fc115da0-eada-11e6-9e39-7b2b3b1f5f66.png)

## Expected

Same behavior as homepage: geolocation with ipinfo.org fallback.

capndesign:
This is working. Thanks @liamdanger!

PR #160, opened 2017-02-06
Add startup action to about view to get issues

rosemarydotworld:
Fix issue #144 by adding the `startup` action to the about view. 

Next step should probably be to clean this up so that the action doesn't need to be individually called on several components. But for now: fixed!


beausmith:
LGTM!

nickoneill:
Yet another case of this :/

PR #162, opened 2017-02-06
.toLocaleString() expects a Number, not a string.

beausmith:
Same fix as previously merged fix in: https://github.com/5calls/5calls/pull/32. 

Looks like it got lost during the restyling. 😜  cc @jameshome


jameshome:
I still wasn't getting commas without that crazy regex, but I don't think I did type conversion to number, either. If this works, I'm all for it.

PR #168, opened 2017-02-07
Update to San Francisco number listed on www.harris.senate.gov

beausmith:
This SF number was listed as the LA number.


ctaymor:
LGTM

PR #169, opened 2017-02-07
Record phone number with outcome

beausmith:
This required a change to the UI…

change-number

Once an outcome is selected, the phone number is sent to the `report` endpoint:

image

cc @josharian

### DIFF
--- a/static/js/components/contact.js
+++ b/static/js/components/contact.js
@@ -6,25 +6,25 @@ module.exports = (c, state, prev, send) => {
   const reason = c.reason == "" ? "This organization is driving legislation related to the issue." : c.reason;
   const repID = c.party ? c.party.substring(0,1) + "-" + c.state : '';
 
-
-  let fieldOffices
-  if (!!~c.field_offices) {
-    fieldOffices = html`
-      <p class="call__contact__show-field-offices">Busy line? <a onclick=${() => {send('toggleFieldOfficeNumbers')}}>Call a local office</a></p>
-    `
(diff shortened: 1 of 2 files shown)


josharian:
Sorry to be a pain, but what if instead of using direct tel hrefs, we used a javascript hook that recorded the last tapped tel and then redirected to the tel href? Then no UI changes required. I guess that doesn't help on the desktop, though. Hmmm.

nickoneill:
I agree with josh here, making some assumptions on how the numbers are tapped is going to give enough insight into which numbers are being called with the hassle of extra UI.

beausmith:
@josharian @nickoneill - Thanks for the feedback. I understand what you're getting at. I would love to have some synchronous conversations before beginning on other non-obvious work to ensure efforts are put to best use. 😃

phoebejaffe:
Personally, I think a dropdown will lead people to not knowing that there are multiple numbers. It seems obvious to us, but a lot of people just won't see it and know that there are more numbers behind it.

This is especially an issue since for a lot of people, this is their first experience calling reps, and it's scary... anything that gets in the way will cause them to drop out.

I'd prefer to have the "UNAVAILABLE" button prompt the user to 
```
Try another number?
[YES]   [SKIP]
```

That 1. gives us signal that the previous number didn't work, and 2. gives the user a clear path to UI for the next number.

nickoneill:
I think we're good here, having settled on the non-dropdown solution for alternate numbers.

PR #172, opened 2017-02-08
Download, cache, and serve rep images

josharian:
It is late, so review skeptically. I will take another look at this tomorrow myself.

I'm not happy with how ugly it is to thread a reference to the photocache everywhere it needs threading. I'd be happy to substitute a global photocache instead.

In general, this should have minimal impact on response times, but the very first time we touch any given rep, it'll be slow. The waitgroup goo helps. It also slows down initial startup a little, but again, doing all the fetches in parallel helps.

See the commit message for further details.

### DIFF
--- a/go/airtable.go
+++ b/go/airtable.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


nickoneill:
Works great! Thanks!

Issue #170, opened 2017-02-07
UTM Parameter breaks site


jonnaro:
When including a UTM Parameter for GA Campaign tracking, the application persists this parameter in the URL and subsequently breaks site navigation.

To reproduce:
1. Go to: https://5calls.org/?utm_campaign=test
2. Click on an issue
3. Subsequent URL will look like: https://5calls.org/?utm_campaign=test#issue/recSWqDICerrEOUDC

This is incorrect, as the site should not persist the utm_campaign parameter in the URL after the first call. And by remaining in the URL, the site navigation does not work.

jonnaro:
Duplicate of #165

PR #174, opened 2017-02-08
Use Choo to change url. Fixes #161

### LINKED ISSUE #161: Complaints from Windows 7 users
Apparently windows 7 doesn't work on the site, issues don't show up.

oderby:
My brief reading of Choo documentation suggests we
should be using Choo's internal location model's reducer
when we want to change the url. Especially since we are
relying on it elsewhere to determine the params. See
https://github.com/yoshuawuyts/choo/blob/f84ec43fa58508cc20fe537d752a14901339f0cd/README.md#router
for more information.

With this change, selecting issues works for me
on IE 11 in Windows 7, fixing #161.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -207,10 +207,6 @@ app.model({
         send('receiveTotals', body, done)
       })
     },
-    changeActiveIssueEffect: (state, issueId, send, done) => {
-      send('location:set', "/#issue/"+issueId, done)
-      send('changeActiveIssue', issueId, done)
-    },
     setLocation: (state, data, send, done) => {
       send('setAddress', data, done);
       send('fetch', {}, done);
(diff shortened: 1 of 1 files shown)


nickoneill:
Hah, that's fantastic. Does this also behave well with urls that have ?utm_source=xxx inserted? This was the workaround mentioned directly below.

oderby:
will check. How much of the "activeIssue" state is old/cruft? AFAICT, the code at https://github.com/5calls/5calls/blob/master/static/js/main.js#L210 is never called. Probably better left to a separate commit, but just noticed it while digging here.

nickoneill:
I didn't realize we had a second effect that wasn't used here, so probably all cruft. I think it makes sense to remove it in this commit.

oderby:
Okay, following the steps outlined in #170 , I can confirm the site behaves the same with this fix as on prod (query param is cleared when you navigate to a site).

I will remove the unused effect, add some comments and push up a new revision for the PR.

oderby:
Also, this might be a good candidate for a regression test once browser-based testing lands in #131 But not at all necessary to get this fix in ;)

Issue #171, opened 2017-02-08
Sometimes contacts from google civic get emptied for a bit


nickoneill:
Every once in a while I'll check on something for a user where they say reps aren't being shown for them and, sure enough, there will be no house or senate reps for that particular zip code.

Restarting the app server fixes things. I wonder if we're getting back weird blips from the civic API sometimes and caching that bad data?

agjohnson:
Any more information on this -- debug logs or perhaps info from the Google API console? A first step here might be to add some more defensive logging.

Issue #64, opened 2017-01-30
Add clarity on why ZIP may not return all your representatives?


flyingrat42:
For "split" zip codes (like mine), the API appears to return only the Senators when only the ZIP is entered, because the full street address is needed to resolve my congressional district.  Can some language be provided on the location bar to clear this up so people don't wonder why they're not getting their Representative's contact info?

nickoneill:
We return a `splitDistrict` value in the JSON, it's just up to the frontend to determine if that's true and prompt for a better locaiton

oderby:
In case it helps, here's some sample copy (taken from [here](http://act.commoncause.org/site/PageNavigator/ElectedOfficials.html))
![split_zip_copy](https://cloud.githubusercontent.com/assets/1351158/22757261/fd2bf64c-edfe-11e6-93ff-f21f2bc5814b.png)
(Note: 94110 is not actually a split zip for federal reps, only at the local level, but you get the idea) It would be great if someone can provide a few examples of "split" zip codes for whoever fixes this to test with...

oderby:
Is anyone working on this issue?

A user ran into this issue today, so recording what was found from that investigation:
Some test cases: 33408, 04103, 94043, 60647

It appears that even with more information, sometimes civic api doesn't give congressional information. An example is 33408-3231, for which [civic api](https://developers.google.com/apis-explorer/#p/civicinfo/v2/civicinfo.representatives.representativeInfoByAddress?address=33408-3231&levels=country&roles=legislatorLowerBody&_h=1&resource=%257B%250A%257D&) still doesn't list a congressional district, but the [government does](http://ziplook.house.gov/htbin/findrep?ZIP=33408-3231&Submit=FIND+YOUR+REP+BY+ZIP). I've submitted a bug report to them about this particular issue.

However, it's quite likely that other users will run into this issue, even if they put in a specific address. So we should consider this possibility when thinking through the messaging around possible failure cases.

charkattack:
I'm getting a lot of emails about the split districts after the Michael Moore mention. Here are a few of them:
- My Zip code has multiple districts. Since your app doesn't go past the zip code level, I'm not sure if I am speaking to the correct representative's office.
- I put in my zip and it will not show my rep. I've in my location and can't see my reps 
- HELP!!!! Please fix the location service to allow those of us overseas or expats to manually record our address. My zip code has multiple representatives, and being overseas, I can't use the location services.
[...]

PR #175, opened 2017-02-09
Different "No Contacts" message if split district (partial fix to #64)

ColemanCollins:
This is a partial fix for #64, which has come up on the slack a few times in the past week. 

Specifically, provides a more helpful message for the most general case in the issue, where we have location resolved as a split district but there are no contacts for a given issue. 

The fact that there's a split district isn't the only possible reason that no contacts could be returned for an issue when they've provided a valid location (see #171, etc), and entering a more specific location isn't necessarily going to fix it (in the case where it's just an address that the Civic API doesn't like), but it seems to be the most likely case AND the only one a user can take action on, so even in those cases it seems like a better message than the generic "set location" message.

It would be ideal to also provide messaging around this within issues, but the API doesn't return any information that would allow us to know that an issue _would have returned a house rep_ if it hadn't been a split district.

That would require some sort of API change to communicate _intended_ contacts (eg: only senators, only house reps, both) as well as the resolved ones, to know if we got what we were looking for (and message appropriately). Opening an issue around that shortly.

### DIFF
--- a/static/js/components/noContact.js
+++ b/static/js/components/noContact.js
@@ -8,8 +8,7 @@ module.exports = (state, prev, send) => {
   }
 
   function noContactsMessage(state) {
-    if (state.splitDistrict == true && (state.address || state.cachedCity)) {
-    // if it's a split district but we have a location set
+    if (state.splitDistrict && (state.address || state.cachedCity)) {
       return html`<div>
                     <p>The location you provided could be in one of two congressional districts.</p>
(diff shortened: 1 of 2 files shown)


nickoneill:
Thanks! Tweaked just the smallest thing here to have a lowercase 'zip'. I'm not sure people will know what that is but hopefully they'll figure it out or just use an address.

PR #179, opened 2017-02-11
Move injection of default contact reason to backend.

oderby:
This avoids the need for every client to implement their own
default behavior.


nickoneill (approved):
👍

PR #134, opened 2017-02-03
Additional JavaScript component unit tests

cdoremus:
This PR adds JavaScript unit tests for half-dozen components as described in issue #131.

PR #181, opened 2017-02-11
Run front-end tests in browsers via Karma

Mr0grog:
Addresses *part* of #131. Sorry this has taken me a week to get to. Have at it; let me know if this is any good.

I looked into a few different solutions (more below), but I think Karma with Mocha is going to be the best way forward here.

- It’s well used with lots of tools and plugins (easy to add browsers, coverage tools, SauceLabs/Browserstack, etc.)
- It handles sourcemaps and browserify great.
- Configuration is mostly straightforward and doesn’t add a lot of complexity to the codebase.

This sets us up with a simple config that runs tests in Chrome. It’s pretty easy to add other browsers, but I think we want to keep basic while-you’re-developing tests simple and fast (no reason for someone volunteering their help to get held up because they don’t have browser XYZ).

You can now also run `gulp test --watch` and have it watch JS files and re-run the tests whenever you make a change. (I noticed after writing this that there are `xxx:watch` tasks; should change this to match those? I like that it’s one implementation instead of two, but it’s inconsistent with everything else.)

There’s [another branch](https://github.com/Mr0grog/5calls/compare/tests-with-karma...Mr0grog:tests-with-karma-ci) where I was able to get tests across many browsers working nicely in Sauce Labs; that should be a next step when we add CI support.

---

### Side Notes

Some other experiments involved:

- [`jsdom-global`](https://github.com/5calls/5calls/compare/master...Mr0grog:localstorage-tests-with-jsdom-global) uses a JS implementation of the DOM to give us a browser-like environment in node. It’s fast and light, but not exceedingly accurate, since it’s not an actual browser engine. This last part makes me not-too-fond of it.
- [`gulp-mocha-phantomjs`](https://github.com/5calls/5calls/compare/master...Mr0grog:tests-with-phantomjs) This only runs tests in PhantomJS and is a bit more work on our side, though it is a little faster and lighter than Karma. I also wasted a stupid amount of time trying to get source maps working well here.
- [`mochify`](https://github.com/5calls/5calls/compare/master...Mr0grog:tests-with-mochify) is halfway in-between Karma and `gulp-mocha-phantomjs`. It’s much more configure-and-go, though getting it to work with browsers besides Phantom and Slimer is complicated (you need to get a webdriver server running), but at least it has support. I also spent a dumb amount of time fixing sourcemaps here, too :\
- [Intern](https://theintern.github.io) is a test runner I’ve been meaning to play with for a while (it supports unit tests and integration tests together in one package), but it turns out to be a little bit of a bear to use with CommonJS (it’s doable, but feels like a hack). I didn’t wind up exploring too far down this route.

Also worth noting: it turns out PhantomJS doesn’t do localization, so `number.toLocalString()` doesn’t do anything different than `number.toString()` in it. That’s actually more-or-less up-to-spec, but it’s not nice and it causes the `callcount` tests to fail. That might say more about `callcount` than it does about Phantom, though.

Anyway, if we choose to use Phantom for any tests, we’ll need to make sure to either skip the `callcount` tests in Phantom or to do some work on that module so that it functions in Phantom.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,7 +12,7 @@ var gulp = require('gulp')
   , uglify = require('gulp-uglify')
   , http_server = require('http-server')
   , connect_logger = require('connect-logger')
-  , mocha = require('gulp-mocha');
+  , spawn = require('child_process').spawn
   ;
 
 var SRC = {
@@ -112,11 +112,28 @@ gulp.task('extra', function() {
     .pipe(gulp.dest(DEST.html));
 });
 
-gulp.task('test', function() {
-  return gulp.src(['**/*_test.js', '!./node_modules/**'], { read: false })
-    .pipe(mocha({
-      reporter: 'spec',
-    }));
+gulp.task('test', function(done) {
+  const karmaArguments = ['start'];
+  if (!process.argv.includes('--watch')) {
+    karmaArguments.push('--single-run');
+  }
+
+  // Karma has a nice public API, but has issues where it can hang when trying to shut down after completing tests, so
+  // run it as a separate process instead. See:
+  // https://github.com/karma-runner/karma/issues/1693
+  // https://github.com/karma-runner/karma/issues/1035
+  const karma = spawn('./node_modules/.bin/karma', karmaArguments, {
+    cwd: process.cwd(),
+    stdio: 'inherit'
+  });
+
+  karma.on('close', code => {
+    if (code) {
+      done(new util.PluginError('Karma', `JS unit tests failed (code ${code})`));
+      return
+    }
+    done();
+  });
 });
 
 gulp.task('default', ['html', 'html:watch', 'html:serve', 'sass', 'sass:watch', 'copy-images', 'copy-images:watch', 'scripts', 'scripts:watch', 'extra']);
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,66 @@
+// Karma configuration
+// Generated on Tue Jul 26 2016 11:30:01 GMT-0700 (PDT)
+
+module.exports = function (config) {
+  config.set({
+     // base path that will be used to resolve all patterns (eg. files, exclude)
+    basePath: '',
+
+    // frameworks to use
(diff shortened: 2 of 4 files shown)


cdoremus:
Nice job Rob (Mr0grog)! I am in awe of your thorough treatment of this subject. I agree that this Karma/Mocha PR is the best option.

@nickoneill, I hope you agree and will promptly merge this PR into master. After that, I will update PR #134 to work with this code.

oderby (commented):
These look good to me - just pulled it down and ran locally and it looks good. We should keep the gulp interface consistent, so probably should refactor to have a second "test:watch" routine. Otherwise seems fine to me. I can't wait to write a test for #161, once we get saucelabs up and running (IE test needed there)

cdoremus:
I too was able to get this Karma configuration to work locally. In addition, I was able to update my unit tests from PR #134 with this configuration and add a few extra component tests.

nickoneill:
I'm totally unfamiliar with this kind of testing so forgive my inexperience. I ran this locally and it test things! Works great! Looks like 12/5 for me but that may have been because of that last merge I did due to #134?

What do you need to get this running for saucelabs? Should I merge this now? Start a saucelabs account for 5 calls? In your court, just let me know.

nickoneill:
Merged at the request of @cdoremus - please let me know here or slack about saucelabs!

Mr0grog:
@nickoneill I don’t have access to your Slack, but yeah, you should create a Sauce Labs account. Open source projects are free. We can also do Browserstack if you are more familiar or comfortable with that—Karma has plugins for both, though I haven’t tested the Browserstack one.

You should also set up an account with CircleCI so we can get tests to automatically run for all the PRs. I’ll work on a PR this week to add configuration for both Sauce (or Browserstack, LMK which) and Circle.

@oderby Was the issue in #161 particular to Windows 7 or does it occur in 8 and 10 as well? Trying to figure out what browser/platform combinations we should set up for CI tests.

Tracking this work in #188.

oderby:
@Mr0grog responded to your question on #188

PR #182, opened 2017-02-11
pass golint

buddhamagnet:
Love this, I'm a freelance hacker with experience of both modern JS (right now working with React and Redux) and Go, so happy to help. First step, ensure the back end code passes lint.


nickoneill:
Thank you! My golint integration in sublime text has been misplaced.

PR #189, opened 2017-02-13
Rename test task to watch & auto-run tests to `test:watch`

Mr0grog:
As noted in #181 by @oderby, there should have been two tasks for tests (`test` and `test:watch`) instead of one task with an optional argument (so it matches the existing gulp tasks). That PR got merged before I had a chance to make that change, hence this separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,13 +107,14 @@ Edit the `appUrl` variable in `static/js/main.js` to be `http://localhost:8090` 
 ## Unit tests
 
 #### JavaScript 
-JavaScript unit tests are written using ```Mocha``` and ```Chai``` and run in the ```Karma``` test runner.
+
+JavaScript unit tests are written using ```Mocha``` and ```Chai``` and run in the ```Karma``` test runner. You must have the Google Chrome browser installed to run them.
 
 Run the unit tests with:
 
 ```npm test```
 
-Run the unit tests in watch mode with:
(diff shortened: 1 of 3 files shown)


Mr0grog:
Oops, I missed that @cdoremus had also created a matching NPM script for `test:watch`. Added a second commit that updates it.

Do we actually want to keep it, though? `npm test` is an official, specially sanctioned script, so it makes sense that it’s there, but since we don’t mirror all the other gulp tasks as NPM scripts, I’m not sure whether it makes sense to keep `npm run test:watch`.

oderby:
I think it's minor. Personally, I like the consistency of just using one tool for all our frontend build/test flows and think this looks fine. If we want to have npm script targets aliasing all gulp tasks, that's fine, but I don't see a reason to only include a subset, and it could become unwieldy after a while. But that's just my $0.02 - would be curious if @cdoremus has any thoughts?

cdoremus:
I'm use to having package.json as the single source for information on building a frontend app. I just think it is a good way to decouple build commands from build mechanisms (using Gulp in our case). Still, I'm not dogmatic about this and am willing to go along with a consensus view. I've added a Unit Tests section to the Readme, so if you are going to make any changes to how tests are run, please update that section.

nickoneill:
It looks like this is up-to-date in the Readme @cdoremus so I'm going to merge.

PR #190, opened 2017-02-13
Fix #185 - focus and select input when prompted for address.

### LINKED ISSUE #185: "set your location" or "add an address" prompts should clear and focus the input
Some feedback with confusion on these items. Clearing the input lets people see the placeholder where it says "address or zip"

oderby:
Adds a new effect 'focusLocation' to achieve this, called
from within issuesLocation component if appropriate.

TODO: how to write tests for this?

### DIFF
--- a/static/js/components/issuesLocation.js
+++ b/static/js/components/issuesLocation.js
@@ -1,6 +1,10 @@
 const html = require('choo/html');
 
 module.exports = (state, prev, send) => {
+  if (state.askingLocation && !state.fetchingLocation) {
+    send('focusLocation');
+  }
+
   return html`
     <div class="issues__location">
     ${pretext(state)}
--- a/static/js/components/issuesLocation.js
+++ b/static/js/components/issuesLocation.js
@@ -49,10 +49,6 @@ module.exports = (state, prev, send) => {
 
   function enterLocation(e) {
     e.preventDefault();
-    // Clear previously invalid address to reinforce entering a new one
-    if (state.invalidAddress) {
-      document.getElementById("address").value = "";
-    }
     send('enterLocation');
   }
 
--- a/static/js/components/issuesLocation.js
+++ b/static/js/components/issuesLocation.js
@@ -15,25 +15,22 @@ module.exports = (state, prev, send) => {
   function pretext(state) {
     if (state.fetchingLocation) {
       return html`<p class="loadingAnimation">Getting your location</p>`;
-    }
-    else if (state.askingLocation) {
+    } else if (state.askingLocation) {
       return html``;
+    } else if (state.invalidAddress) {
+      return html`<p><button class="subtle-button" onclick=${enterLocation}>That address is invalid, please try again</button></p>`;
+    } else if (state.address) {
+      return html`<p>for <button class="subtle-button" onclick=${enterLocation}>${state.address}</button></p>`;
+    } else if (state.cachedCity) {
+      return html`<p>for <button class="subtle-button" onclick=${enterLocation}> ${state.cachedCity}</button> ${debugText(state.debug)}</p>`;
     } else {
-      if (state.invalidAddress) {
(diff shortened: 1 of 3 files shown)


oderby:
First, I'm not entirely sure this is the "choo" idiomatic way to achieve this - basically we want to modify the DOM immediately after it's rendered/updated. It seems like the best place for this logic is in an effect...but I'm open to other ideas? Couldn't really find any good examples of recommended ways to do this within Choo's docs.

Also, @Mr0grog @cdoremus I'd love to hear if you have any suggestions for ways to test this sort of thing. In general, the challenge is testing the effects of DOM manipulations carried out in model effects. This is probably not a simple unit test, as it's testing multiple parts of the app working in concert. Did either of you find any canonical examples/resources for testing Choo applications? All I've found was [this presentation](https://www.sitepoint.com/functional-programming-choo/), but it only talks about unit testing views/components or reducers.

cdoremus:
I've put together some tests for your new issuesLocation code, but there is a lot of other logic in that component that you should test:
```
describe('issuesLocation component', () => {
  it('should call send("focusLocation") when askingLocation is true and ' +
    'fetchingLocation is false', () => {
    const expected = 'focusLocation';
    let actual = undefined;
    // mock impl to test that it is being called in component
    let send = (msg) => actual = msg;
    let state = {
      askingLocation: true,
      fetchingLocation: false
    };
    issuesLocation(state, null, send);
    expect(actual).to.equal(expected);
  });
  it('should NOT call send("focusLocation") when askingLocation is true and ' +
    'fetchingLocation is true', () => {
    const expected = 'focusLocation';
    let actual = undefined;
    let send = (msg) => actual = msg;
    let state = {
      askingLocation: true,
      fetchingLocation: true
    };
    issuesLocation(state, null, send);
    expect(actual).to.not.equal(expected);
  });
  it('should NOT call send("focusLocation") when askingLocation is false and ' +
    'fetchingLocation is false', () => {
    const expected = 'focusLocation';
    let actual = undefined;
    let send = (msg) => actual = msg;
    let state = {
      askingLocation: false,
      fetchingLocation: false
    };
    issuesLocation(state, null, send);
    expect(actual).to.not.equal(expected);
  });
});
```
The focusLocation effect in main.js cannot be tested unless the function is exported. Personally, I think that main.js is getting too long and complicated. I'd recommend that the state, reducers and effects be  refactored into separate files to allow easier testing and better code clarity.

oderby:
@capndesign any further thoughts on this PR? (Asking because you requested to review.) I'm inclined to merge as is and track coding better tests in a separate issue.

cdoremus:
As I stated above, the tests I provided covered a very small subset of the issuesLocation component logic. I would like an issue created to fill in the coverage gaps before this PR is merged into master. Testing is always better done close to the initial development of a new piece of code as possible. I've seen too much testing procrastination in my days as a dev (over 20 years) that ends up being wishful thinking. We've already derived value from our unit tests. This value will disproportionately increase as our code coverage increases.

oderby:
Good point. I will try and get something up with better unit test coverage
for the issuesLocation component soon. Hopefully tonight.

oderby:
@cdoremus @capndesign either of you up to review this? I think it's ready to merge otherwise...

cdoremus:
@oderby, I'll do the review and try to get to it tonight

cdoremus:
I guess that I do not have 'reviewer' status to do an official review, but I did take a look at the code, paying special attention to the tests, and see no reason why it cannot be merged into master. Good job!

nickoneill:
Is there a "review changes" button if you swap to the files changes tab @cdoremus ?

capndesign (commented):
Everything looks good in the diff and the code runs as expected. I’m having trouble getting the karma tests to run on this machine, but other than that things look good to me.

cdoremus (approved):
Code looks good. All tests pass. This branch is ready to merge to master. Good job!

nickoneill:
👍

PR #192, opened 2017-02-14
Fix function names

blip-lorist:
Noticed while setting up that I was getting errors while running `go get` and `go build.` Looks like this function was recently renamed, just updated a few uses of it.


nickoneill:
golinted but didn't check that it was building correctly, shame on both me and @buddhamagnet :P

PR #193, opened 2017-02-15
Add Continuous Integration with CircleCI

Mr0grog:
This adds configuration and support for running the JS unit tests in CircleCI. The tests also use Sauce Labs to run simultaneously in a variety of browsers. Instead of running `npm test` as normal, the CI server runs `gulp test:continuous-integration`, which uses the `karma.ci.conf.js` configuration file, which uses Sauce Labs to run tests on lots of browsers.

## Setup

This is just the code and configuration. Accounts need to be set up in order for all this machinery to run when someone pushes a commit to 5calls. The steps here are the same for the main repo and for others who want to run CI and Sauce on their own in their forks (I think most folks shouldn’t to worry about this since PRs should automatically run on the main repo’s account).

### Set up Sauce

1. Create an account at https://saucelabs.com
2. Log in and click on your username in the lower left corner, then “my account,” then scroll down to “access key” and click to show your key. Copy it down (you’ll need it to configure Circle).

  get-access-key

3. That’s it!

### Set up CircleCI

1. Create an account at https://circleci.com
2. Log into Circle and…
    1. Click on “add projects” in the sidebar, then
    2. Find the Github org/user that owns the repo you want to turn on CI for (in this case, `5calls` for the official repo or your username for your fork), then
    3. Find the `5calls` repo and click “build project.” This will tell Circle to start automatically running builds/tests for every push and PR.

      set-up-circle

    4. Once that’s set up, you should see the build screen! Scroll down and watch as each step in the build process runs. (It’ll go faster on future runs, when dependencies and environments are cached from previous tests.)

    5. Now click on the gear in the top right to configure the environment (we need to add login credentials for Sauce).

      Configuration Button for CircleCI Builds

    6. Click on “environment variables” on the left, then click the “add variable” button. Do this twice to add two variables:
        - `SAUCE_USERNAME`, which should have your username on Sauce Labs (it’s case-sensitive) for the value.
        - `SAUCE_ACCESS_KEY`, which should have your access key from Sauce (above) for the value
      
        (We use set up the environment variables this way because the credentials are secret and we don’t want them in the git repo itself.)

        Add Environment Variables on Circle

Back in Github, Circle should have automatically integrated itself. On any PR where tests are running, you’ll see a nice notification about success or failure right above the merge button. Click “details” to go to the build page on circle from before:

test results visible next to merge button in github


Mr0grog:
D’oh, forgot instructions on setting up Sauce credentials on Circle. Just updated the PR description—you’ll have to check it on Github if you are reading via e-mail.

Mr0grog:
Side notes—you’ll notice the build actually fails right now in some browsers:

- Safari 9 fails the `callcount` test for the same reason Phantom JS does (described more in #181). Basically, it doesn’t have number formatting based on locale (that was added in Safari 10).
- IE 10 fails because of `console.debug` calls. IE is the worst offender here, but some browsers either don’t create a `console` object until the debugger is opened or don’t have methods other than `log()` on the `console` object. This is why a lot of projects use a logger like [`loglevel`](https://www.npmjs.com/package/loglevel) or [others](https://www.npmjs.com/search?q=log&page=1&ranking=optimal)

We can:

- Disable those tests for now
- Disable those browsers for now

oderby:
I'd vote to disable the offending browsers (assuming just IE 10 and Safari 9) for now, and create a issue to get them working.

cdoremus:
+1 @oderby 

@Mr0grog, if the CircleCI/Sauce Labs tests are run with every PR, is there a way to run these tests locally before the PR is pushed to this repository? If not, a lot of trial-and-error PR commits could be required to get some tests to pass.

In other words, will ```gulp test:continuous-integration``` run locally?

gulpfile.js
```diff
@@ -140,11 +145,15 @@ function runKarmaTests (singleRun) {
...
-  return runKarmaTests(true);
+  return runKarmaTests({singleRun: true});
 });
 
 gulp.task('test:watch', function() {
-  return runKarmaTests(false);
+  return runKarmaTests();
+});
+
+gulp.task('test:continuous-integration', function() {
```
oderby, line 155: Can we just make this 'test:ci' ?

Mr0grog:
> is there a way to run these tests locally before the PR is pushed to this repository?

Sure—just do the same thing as the CI server: run `gulp test:continuous-integration` (you’ll need a Sauce account, of course).

If you’re working and want to debug interactively in a local browser that’s not Chrome, you can always `gulp test:watch`, then open any browser you like and browse to `http://localhost:9876` (or on another computer in your local network, `http://dev.machine.ip.address.here:9876`). And obviously modify the configs to run other browsers automatically instead of manually.

oderby:
As a clarification point - I believe anyone can create their own free saucelabs account (it might only accept tests from a specific public repository, unclear), thus getting their own access_key to use locally.

And to add a note regarding process here - as @cdoremus points out, PRs should should definitely pass tests on chrome locally before being submitted. I don't think we should expect developers to need to run tests on any other browser or on saucelabs before submitting - that's the whole point of setting up CI. 

While it's possible PR authors will run into issues with tests that fail on IE 11 or such on sauce labs, I personally don't think this will be very common situation. And likely if they do, it may or may not be something they're capable of solving (could be they triggered a regression on some obscure bug that only affects Windows) and so would want help/involvement from older/core contributers.

We could consider (securely) distributing the saucelabs accesskey to core frontend contributers if we think that's worthwhile. But I'd recommend not bothering until it becomes a problem.

Mr0grog:
> I believe anyone can create their own free saucelabs account (it might only accept tests from a specific public repository, unclear), thus getting their own access_key to use locally.

Yes (and yes, once the trial period is over, the only free tests are from public repos, but if they have created a fork in order to submit a Github PR, their fork will probably be public).

> I don't think we should expect developers to need to run tests on any other browser or on saucelabs before submitting - that's the whole point of setting up CI.

Totally agree. Sorry I didn’t think to call that out specifically.

> We could consider (securely) distributing the saucelabs accesskey to core frontend contributers if we think that's worthwhile.

I don’t think it really makes sense to do this (see point above about forks being public).

nickoneill:
I think I configured this correctly, merging to test.

nickoneill:
I'm genuinely confused why this works and how all the glue knows what to do! But I'm happy it does, see this PR for the first tests run automatically: #196 

And these two relatedly: #197, #198

Mr0grog:
Whoa! I was going to change the task name (per @oderby) and turn off the failing tests. I’ll submit separate PRs for those ASAP.

PR #199, opened 2017-02-16
Fix gulp task 'test:watch' to work.

oderby:
Just a tiny error made the "test:watch" task break. Must of occurred with all the shuffling that happened to land the CI PR.


Mr0grog:
Heh, thanks. Didn’t see this, so it’s doubly fixed with #200 :P

PR #200, opened 2017-02-16
Fix CI test failures (follow-on from #193)

Mr0grog:
Fixes a bunch of things I didn’t get a chance to remedy before #193 was merged.

- Fix `test:watch` task
- Change `test:continuous-integration` task to `test:ci`
- Don’t run `callcount` tests on browsers without number formatting
- Disable IE 10 in CI tests for now (because it always fails)

PR #202, opened 2017-02-16
Fix support check that skips tests on ALL browsers

Mr0grog:
Mistake and poor checking on my part in #200.

@nickoneill No offense, but maybe it would be good to have a policy of requiring at least one reviewer with a clear understanding of what the code is doing give a 👍 before merging? Would help mitigate dumb mistakes like this.

### DIFF
--- a/static/js/components/callcount_test.js
+++ b/static/js/components/callcount_test.js
@@ -4,12 +4,12 @@ const chai = require('chai');
 const expect = chai.expect;
 
 // Skip a test if the browser does not support locale-based number formatting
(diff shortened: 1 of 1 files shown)


nickoneill:
Yeah, trying to walk the line between being fast enough so people don't feel ignored and finding a volunteer to do a code review in a reasonable time. I'll start assigning some reviews and see how fast people get to it!

oderby (approved):
lgtm

nickoneill:
Changes approved, checks passed

Issue #201, opened 2017-02-16
Add `console.*` abstraction (to support IE < 11)


Mr0grog:
This is mainly about support IE versions 10 and earlier. They don’t create a `console` object until the dev tools are opened, which means the `console.debug()` calls in the JS code throw exceptions and stop the site from functioning in many versions of IE. Not all browsers support the same set of console methods, either.

We should add an abstraction for console logging to handle this. It can be as simple as a module with a several functions like:

```js
module.exports = {
  log (...args) {
    if (window.console && console.log) {
      console.log(...args);
    }
  }
}
```

There are also plenty of modules available on NPM for this. As a good, simple, browser-focused one, I’d recommend [loglevel](https://www.npmjs.com/package/loglevel).

PR #203, opened 2017-02-16
Use loglevel module for JS logging

Mr0grog:
This addresses #201 and allows us to turn IE 10 back on for testing.

There are other ways to handle this and modules we could use besides [loglevel](https://www.npmjs.com/package/loglevel), but I think it’s a good, simple one (and I have experience with it). Happy to close this is others prefer something else.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -144,7 +144,7 @@ function runKarmaTests ({singleRun, configFile} = {}) {
   });
 }
 
-gulp.task('test', function() {
+gulp.task('test:js-unit', function() {
   return runKarmaTests({singleRun: true});
 });
 
@@ -155,9 +155,32 @@ gulp.task('test:watch', function() {
 // Designed for running tests in continuous integration. The main difference
 // here is that browser tests are run across a gamut of browsers/platforms via
 // Sauce Labs instead of just a few locally.
-gulp.task('test:ci', function() {
+gulp.task('test:ci', ['eslint'], function() {
   return runKarmaTests({configFile: 'karma.ci.conf.js'});
 });
+
+gulp.task('eslint', function() {
+  const eslint = require('eslint');
+  const linter = new eslint.CLIEngine();
+  const report = linter.executeOnFiles(['./static']);
+  
+  // customize messages to be a little more helpful/friendly
+  report.results.forEach(result => {
+    result.messages.forEach(message => {
+      if (message.ruleId === 'no-console') {
+        message.message = 'Please use the `loglevel` module for logging instead of the browser `console` object';
+      }
+    });
+  });
+
+  process.stdout.write(linter.getFormatter()(report.results));
+  if (report.errorCount) {
+    return Promise.reject(
+      new util.PluginError('ESLint', 'Found problems with JS coding style.'));
+  }
+});
+
+gulp.task('test', ['eslint', 'test:js-unit']);
 
 gulp.task('default', ['html', 'html:watch', 'html:serve', 'sass', 'sass:watch', 'copy-images', 'copy-images:watch', 'scripts', 'scripts:watch', 'extra']);
 gulp.task('deploy', ['html', 'sass', 'build-scripts', 'extra', 'copy-images']);
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -175,8 +175,7 @@ gulp.task('eslint', function() {
 
   process.stdout.write(linter.getFormatter()(report.results));
   if (report.errorCount) {
-    return Promise.reject(
-      new util.PluginError('ESLint', 'Found problems with JS coding style.'));
+    throw new util.PluginError('ESLint', 'Found problems with JS coding style.');
   }
 });
 
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -14,8 +14,9 @@ const appURL = 'https://5calls.org';
 // use localStorage directly to set this value *before* bootstrapping the app.
 const debug = (localStorage['org.5calls.debug'] === 'true');
 
-if (!debug) {
-  logger.setDefaultLevel(logger.levels.WARN);
+if (debug) {
+  // we don't need loglevel's built-in persistence; we do it ourselves above ^
+  logger.setLevel(logger.levels.TRACE, false);
 }
 
 // get the stored zip location
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -2,6 +2,7 @@ const choo = require('choo');
 const html = require('choo/html');
 const http = require('xhr');
 const find = require('lodash/find');
+const logger = require('loglevel');
 const queryString = require('query-string');
 const store = require('./utils/localstorage.js');
 const scrollIntoView = require('scroll-into-view');
@@ -14,7 +15,7 @@ const appURL = 'https://5calls.org';
 const debug = (localStorage['org.5calls.debug'] === 'true');
 
 if (!debug) {
-  console.debug = () => {};
+  logger.setDefaultLevel(logger.levels.WARN);
 }
 
 // get the stored zip location
@@ -29,7 +30,7 @@ store.getAll('org.5calls.location', (location) => {
 cachedGeo = '';
 store.getAll('org.5calls.geolocation', (geo) => {
   if (geo.length > 0) {
-    console.debug("geo get", geo[0]);
+    logger.debug("geo get", geo[0]);
     cachedGeo = geo[0]
   }
 });
@@ -40,7 +41,7 @@ let cachedFetchingLocation = (cachedGeo === '') ? true : false;
 cachedAllowBrowserGeo = true;
 store.getAll('org.5calls.allow_geolocation', (allowGeo) => {
   if (allowGeo.length > 0) {
-    console.debug("allowGeo get", allowGeo[0]);
+    logger.debug("allowGeo get", allowGeo[0]);
     cachedAllowBrowserGeo = allowGeo[0]
   }
 });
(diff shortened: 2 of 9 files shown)


oderby:
How does this interact with logLevel setting in karma.conf? Does karma do it's own override of console? I'm just thinking if there's anything we should do to prevent dumb test failures on CI if someone leaves a `console.debug` call in a test file...

Mr0grog:
Oh, I did not think about how this would interact with Karma’s settings. I *believe* it would just be an additional restriction on top of what Karma does (so whichever is stricter gets applied) because I think it’s just restricting whether it repeats what it gets from the browser, but I’ll have to test and see.

This also made me think we should explicitly put the app into debug mode when running tests (as far as I can tell from perusing the codebase, you have to do this yourself by manually setting localStorage right now).

> I'm just thinking if there's anything we should do to prevent dumb test failures on CI if someone leaves a console.debug call in a test file

Hmm, I think we *want* that to fail, though, right? We should just make sure it fails locally in Chrome, too. I’m not sure if there’s a good way to modify the `console` object to throw errors without stepping on loglevel’s toes, but it does have a plugin system we could potentially use to do it.

That said, the simplest thing might be to add ESLint as part of testing and make it throw errors on `console.*` calls.

oderby:
> We should just make sure it fails locally in Chrome, too.

Yeah, that's probably the only viable approach.

> That said, the simplest thing might be to add ESLint as part of testing and make it throw errors on console.* calls.

To clarify, you're proposing we make `console.*` calls into errors, because people should use `logLevel` instead, right? (When I first read it, I thought you were proposing we make it an error to do any logging from tests) Assuming so, I like the idea. Seems like it's a [built-in rule already](http://eslint.org/docs/rules/no-console) in ESLint. We should also turn it on for app code, too, right? Finally, it'd be awesome if we could customize the error message/give tips to offenders, like "don't use console, use logLevel instead.".

Mr0grog:
> To clarify, you're proposing we make `console.*` calls into errors, because people should use logLevel instead, right? …Seems like it's a built-in rule already in ESLint. We should also turn it on for app code, too, right?

Yep, exactly what I was thinking. And better to just lean on ESLint for this than to somehow make `console.*` calls throw errors, which was what I had initially been thinking.

> Finally, it'd be awesome if we could customize the error message/give tips to offenders, like "don't use console, use logLevel instead."

Good idea. I’m not sure it’s very easy to customize ESLint’s output this way, but I can think of some methods to hack it in.

Mr0grog:
Updated with a couple commits to make sure we always show all logged messages (not just `warn` and `error`) in tests. As a side note, I looked into the Karma config file and:

- `logLevel` only affects messages coming from Karma itself
- `browserConsoleLogOptions` affects which logs from the browser are passed along by manipulating the `console` object in the browser, but does so in a way that doesn’t conflict with `loglevel`. It only works as an additional level filter on top of what we do with `loglevel`.

I also added in ESLint (run `gulp eslint`), but only using the rule for `console.*` calls (it would be good to add more rules later). The message has been customized to:

```
Please use the `loglevel` module for logging instead of the browser `console` object
```

This makes the implementation look a little complicated, but we still use all the standard options and behave as if you just ran `eslint ./static` from the command line (excepting the customized messages).

The linter also runs as part of the normal `test` task. I think we definitely want that for CI; it seemed like it might be good for local dev, too, but we can split them up again if not.

nickoneill:
Who's best to review here? @oderby or @cdoremus?

gulpfile.js
```diff
@@ -144,7 +144,7 @@ function runKarmaTests ({singleRun, configFile} = {}) {
   });
 }
 
-gulp.task('test', function() {
+gulp.task('test:js-unit', function() {
```
oderby, line 147: why not just

```javascript
gulp.task('test' ['eslint'], function() {
```

and then no need for line 183?

oderby: Re-reading your comment on the PR thread, I'm guessing you were thinking this way, people who really didn't want to run ESLint for local dev don't have to. But I'd argue that local tests should be subject to same restrictions/requirements as CI, so with the current config (only run no-console rule), it should definitely run for all local tests.

oderby: Also, as the code currently is arranged, if I run `gulp test`, the two tasks (`eslint` and `test:js-unit`) will run in parallel. This means that eslint could fail while js-unit passes, but because js-unit takes longer, my console is filled with the karma reporter output and I don't notice that the linter failed. The process does correctly return a non-zero exit code, but I doubt most would notice that.

Whereas, if you make the js-unit task depend explicitly on the eslint task (like the test:ci task), it sees eslint fails and doesn't even run karma (much better behavior, IMO)

gulpfile.js
```diff
@@ -155,9 +155,32 @@ gulp.task('test:watch', function() {
...
+      }
+    });
+  });
+
+  process.stdout.write(linter.getFormatter()(report.results));
+  if (report.errorCount) {
+    return Promise.reject(
+      new util.PluginError('ESLint', 'Found problems with JS coding style.'));
+  }
+});
```
oderby: Shouldn't you return a resolved promise here? (empirically, it seems like gulp works fine without it, but for consistency and correctness, seems wrong to only return a promise on failure)

static/js/main.js
```diff
@@ -14,7 +15,7 @@ const appURL = 'https://5calls.org';
 const debug = (localStorage['org.5calls.debug'] === 'true');
 
 if (!debug) {
-  console.debug = () => {};
+  logger.setDefaultLevel(logger.levels.WARN);
 }
```
oderby: It appears (at least for me) the default log level is already WARN. That means that we need to explicitly set the log level to something lower (TRACE) if debug is true. Otherwise, turning on debugging actually has no effect.

Also, did you decide to use `setDefaultLevel` here and `setLevel(level, false)` for the test setup?

static/test/setupUnitTests.js
```diff
@@ -0,0 +1,7 @@
+/**
```
oderby, line 1: This seems reasonable to me, but @cdoremus may have thoughts?

oderby (changes requested):

gulpfile.js
Mr0grog: > I'm guessing you were thinking this way, people who really didn't want to run ESLint for local dev don't have to.

Nah, I was actually thinking more about things like #209 (which I have not had time to look at yet). Mainly this: there will be more “tests” than just the unit tests and the `test` task should just be gathering them all together to run—each type of test should be implemented separately.

> eslint could fail while js-unit passes, but because js-unit takes longer, my console is filled with the karma reporter output and I don't notice that the linter failed.

Sure. I figured it was OK enough that gulp does the job of keeping the output separate, whatever order it was that it came in. We could certainly make them serial and stop as soon as one fails.

static/js/main.js
Mr0grog: You’re right! It’s been a while since I’ve used `loglevel`—could have sworn it was the other way around (trace by default).

> Also, did you decide to use setDefaultLevel here and setLevel(level, false) for the test setup?

Yes, that’s intentional; the two behave differently (`setDefaultLevel` only sets if no level has been explicitly set before, e.g. on a previous page load, while `setLevel(whatever, false)` *always* sets the level, but does not save it). This ensures that, in tests, we are absolutely always showing all logs, but not in a way that might affect your browser next time you open a page not in a test setting.

gulpfile.js
Mr0grog: Solved this by just throwing the plugin error, which is what I should have originally been doing because it’s a sync task.

static/js/main.js
Mr0grog: Done.

gulpfile.js
Mr0grog: Conveniently solved by throwing in https://github.com/5calls/5calls/pull/203#discussion_r103061645 !

gulpfile.js
```diff
@@ -155,9 +155,31 @@ gulp.task('test:watch', function() {
...
 // Sauce Labs instead of just a few locally.
-gulp.task('test:ci', function() {
+gulp.task('test:ci', ['eslint'], function() {
   return runKarmaTests({configFile: 'karma.ci.conf.js'});
 });
 
+gulp.task('eslint', function() {
+  const eslint = require('eslint');
+  const linter = new eslint.CLIEngine();
+  const report = linter.executeOnFiles(['./static']);
```
oderby, line 165: So currently, eslint is run over all files, but only run as part of tests. I believe we should run eslint during normal dev iterations, probably as a dependency of the 'scripts' task? I'm fine filing as an issue and deferring to a separate PR if you'd rather, since the current form at least ensures it will be caught by CI eventually.

oderby (approved):

gulpfile.js
Mr0grog: I don’t see linting as something that belongs as a dependency of the `scripts` task. I don’t believe it makes sense to prevent code from compiling and running simply because it’s poor form; I’ll often try an initial pass at something that is ugly and ignores a bunch of good conventions or rules to test whether it works; making it nice happens later (hopefully before committing, but definitely before PR-ing).

I absolutely think not passing a lint test should prevent code from merging, but by the same token, I see it as an acceptance criterion (tests) and not a build criterion.

oderby:
That's a good point. I was just thinking of "check that they aren't using
console.*" as what would be checked on build, but forgot that it's part of
a larger (planned to expand) lint process. Given that, I'm fine with this
as is.

Issue #207, opened 2017-02-18
Add Selenium end-to-end tests


cdoremus:
I propose adding end-to-end (e2e) testing to the web application using Selenium as an integration test harness. Tests will be written using the Selenium Webdriver JavaScript API in Mocha and Chai.

I have put together a branch containing a working prototype with a simple e2e test to get things going. The tests have been configured to be executed with the web application running locally. If we are going to want to run the e2e tests in CircleCI or Sauce Labs, I will need some help.

My prototype has end-to-end tests in a ```e2e-tests``` folder at the project root.

Comments/suggestions are welcome.

oderby:
I've only had limited experience writing e2e tests for the browser, but in those cases we've used selenium with some success, and I know we've gotten it working with sauce labs, so it's at least possible. So, I'm supportive of using selenium. I expect @Mr0grog might have more to say on this topic though.

Also, @cdoremus do you want to push up your branch so others can review and chime in?

Mr0grog:
👍 Selenium is generally the way I’d go with this. @cdoremus it looks like you may not have pushed your branch up; I don’t see it here or in your fork on Github.

Since I can’t see your branch I don’t know exactly how you’ve set it up, but have you used [webdriverio](http://webdriver.io)? It is a really nice wrapper around the somewhat-painful Selenium Webdriver API. I’ve generally had good experience with it and it includes support for managing tests against an array of browsers pretty much out of the box (via Sauce or any old Selenium Grid server).

cdoremus:
@oderby and @Mr0grog, thanks for your feedback. I had a PR almost ready to go with npm scripts to boot up the front and back end before the tests were run, but they need to be changed now that the back end is in a private repo. I will talk to @nickoneill about how to handle this because everyone who wants to run these Selenium tests will need to be able to run the back end.

@Mr0grog, I just used the generic Selenium Webdriver API. We can discuss changing it to webdriverio after I push the PR.

oderby:
I think with the change, we should actually consider not depending on the
dev version of backend, and just running code against deployed backend. I
generally find this annoying, as your tests can break underneath you due to
changes in deployed server that aren't communicated/deployed, but given
that this is a volunteer project and not under any real time crunch, this
would actually be an effective canary, in that we'd see our PRs/CI fail if
a breaking change is pushed to production without fixing it in the frontend.

nickoneill:
[...]

Issue #206, opened 2017-02-16
Refactor main.js into namespaced models; add tests for effects/reducers


oderby:
As per discussion in #190, we should be able to add unit tests for effects and reducers. Before we can do this, I think it would make sense to break apart the monolithic model in main.js. I think we can do this simply by creating 2 namespaced models, `issues` and `address` (to avoid confusion with the special `location` namespace choo provides for interacting with the `window.location` object), and then just keep a barebones set of initialization logic in the top level model. 

Once that is done, we can start filling in unit tests for our reducers and effects.

nickoneill:
I know this exists in choo but I hadn't yet wrapped my head around how to do it, happy to entertain anything that starts to simplify our gigantic main.js

cdoremus:
I was going to propose that we split state, reducers and effects into their own modules (source files) to improve testability and reduce complexity. Thoughts?

oderby:
In my head, I was envisioning we'd have 1 file per model, which held all state, reducers and effects for that model. I'd imagine we could structure each module such that the model parts are all accessible for testing. Putting everything related to a model in one place makes it much easier to reason about and understand.

@cdoremus how is this different/similar to what you're proposing?

cdoremus:
Owen (@oderby), I agree with your thinking. But I thought some more about it and took a look at ```models.go``` and ```main.js``` and think that a ```contact``` model should be added to the mix.

I'm sorry, but I was confused by your term "namedspaced models". I assume that they are the same as ES6 or Node modules.

greggraf:
I am exploring writing unit tests for `startup()`  which is currently defined in `main.js`.

To  help expose `startup()` for unit testing (and as an incremental step towards refactoring main.js) I would like to propose the following:

- create a file called `./static/js/app.js` (or `application.js` or `5calls.js`)
- move `appURL` constant into this file 
- move all lines of code that interact with `localstorage` to set initial `state` into this file
- this file will be a module that exports two properties: `model`, `routes`
    - `model` will be the object currently passed to `app.model()` in `main.js`
    - `routes` will be the array passed as the second argument to `app.router()
- `require` this file in at the top of `main.js`
- use the exposed properties as arguments to `app.model()` and `app.router()`

This seems pretty safe as it is re-arranging code, not changing the code much. Any issues with variable scope will be pretty quickly exposed as JS errors. 

This module can then be used for unit testing of difficult to reach code in `main.js.`

[...]

PR #209, opened 2017-02-20
Selenium end-to-end integration test setup

cdoremus:
This PR provides end-to-end (e2e) test configuration and an example e2e test as noted in issue #207. Testing uses ```Selenium``` with ```Mocha``` and ```Chai```.

The example is relatively simple display of the 5calls web application's home page and assertion that the page title is correct. The e2e tests are designed to be run locally. They are best used as a sanity check on significant refactorings such as those suggested in issue #206. The configuration will need to be modified  for Sauce Labs integration.

Documentation on how to run the e2e tests has been added to the Readme fille.


package.json
```diff
@@ -5,7 +5,8 @@
   "main": "gulpfile.js",
   "scripts": {
     "test": "gulp test",
-    "test:watch": "gulp test:watch"
+    "test:watch": "gulp test:watch",
+    "e2e": "mocha ./e2e-tests/*.js"
```
Mr0grog: If gulp is our task runner, this should absolutely be managed as a gulp task. It’s OK to have an NPM pass-through, but I don’t think it’s useful—it’s duplicative and doesn’t really add anything. (The exception here are the standardized, particular-to-NPM itself scripts, like `test`, `post-install`, etc.)

e2e-tests/home_e2e.js
```diff
@@ -0,0 +1,53 @@
...
+
+test.describe('home page', () => {
+
+  test.beforeEach(() =>{
+    driver.get(url);
+  });
+
+  test.it('should show correct page title', (done) => {
+    let expected = '5 Calls: Make your voice heard';
+    return driver.getTitle().then((title)=> {
```
Mr0grog: This test should just return a promise rather than both returning a promise *and* accepting a callback function.

Also, spacing between args and arrow: `(title)=>`

e2e-tests/home_e2e.js
```diff
@@ -0,0 +1,53 @@
...
+
+test.after(() => {
+  if (driver) {
+    driver.quit();
+  }
+});
+
+test.describe('home page', () => {
+
+  test.beforeEach(() =>{
```
Mr0grog: Spaces in `=>{`

e2e-tests/home_e2e.js
```diff
@@ -0,0 +1,53 @@
...
+            }
+          }
+      }).build();
+});
+
+test.after(() => {
+  if (driver) {
+    driver.quit();
+  }
+});
```
Mr0grog: These `before`/`after` calls should really be either:

- Inside the `describe` block
- In a separate test setup module (like https://github.com/5calls/5calls/blob/master/static/test/setupUnitTests.js for unit tests) since they cover *all* tests, regardless of the file they’re in.

e2e-tests/home_e2e.js
```diff
@@ -0,0 +1,53 @@
...
+const until = webdriver.until;
+
+const url = 'http://localhost:8000/';
+
+let driver;
+
+test.before(() => {
+  driver = new webdriver.Builder()
+    .withCapabilities({
+          'browserName': 'chrome',
```
Mr0grog: Over-indented by four spaces here?

e2e-tests/home_e2e.js
```diff
@@ -0,0 +1,53 @@
...
+          'timeout': '20000',
+          'chromeOptions': {
+            prefs: {
+              'profile.managed_default_content_settings.notifications': 2
+            }
+          }
+      }).build();
```
Mr0grog: `.build()` should be a new line.

e2e-tests/home_e2e.js
```diff
@@ -0,0 +1,53 @@
...
+ *    gulp
+ * 2. In a second command window:
+ *    npm run e2e
+ */
+const webdriver = require('selenium-webdriver');
+const test = require('selenium-webdriver/testing');
+const chai = require('chai');
+const expect = chai.expect;
+const By = webdriver.By;
+const until = webdriver.until;
```
Mr0grog: `By` and `until` are not used.

Mr0grog (changes requested):
Do we need a `_e2e` suffix on all tests here? Seems to me like being in the `e2e` directory should be enough?

Sorry this has taken me some time to get to.

e2e-tests/home_e2e.js
cdoremus: Is there an eslint config file that you can attach to this issue so I can pre-lint any of my changes?

Mr0grog: No, but for this particular issue, the `newline-per-chained-call` rule is probably what you’d want, e.g:

```json
"rules": {
  "no-console": "warn"
}
```

…but I’m not in love with that rule as it can’t really distinguish (or at least didn’t used to) cases like:

```js
// don't want a warning
thing.something().property;
// do want a warning
thing
  .something({
    stuffAcrossManyLines: true
  })
  .property;
```

package.json
cdoremus: @Mr0grog I tried playing with a Gulp task to do this with no luck, so I'll need your help. Sorry, but I haven't worked with Gulp for a while.

cdoremus:
@Mr0grog I wanted to try out webdriverio because the webdriverjs documentation is sparse, but, according to their docs, webdriverio needs a local selenium standalone server and gekodriver installs. And then you need to run the standalone server separately. This seems like a lot of work for a local e2e test run. There must be an easier way? Please advise.

package.json
Mr0grog: I think it would just be something like we had before:

```js
// you'll need to bring this require call back, since it got removed when we switched to Karma for unit tests
const mocha = require('gulp-mocha');
...
gulp.task('e2e-test', function () {
  return gulp.src('./e2e-tests/*.js')
    .pipe(mocha({reporter: 'spec'}));
});
```

Mr0grog:
> according to their docs, webdriverio needs a local selenium standalone server and gekodriver installs. And then you need to run the standalone server separately.

It’s been a while since I’ve tried to use Chromedriver directly with webdriverio, but I thought I remembered it just working. Otherwise, using the `selenium-standalone` service is the easiest way to get selenium running, I think: http://webdriver.io/guide/services/selenium-standalone.html

Mr0grog:
I don’t want to push over-much on `webdriverio`, though. The `selenium-webdriver` looks much nicer than I remember it being a couple years ago, which is probably the last time I tried using it directly. I don’t know that there’s anything wrong with how you’re using it here and, as you are experiencing, `webdriverio` requires a *lot* more setup.

package.json
cdoremus: @Mr0grog Thanks. This script worked, but I needed to add some setup so the chromedriver could be found in the PATH.

e2e-tests/home_e2e.js
cdoremus: These have been removed

e2e-tests/home_e2e.js
cdoremus: Fixed

e2e-tests/home_e2e.js
cdoremus: Moved .build() to a new line.

e2e-tests/home_e2e.js
cdoremus: Created a setupEndToEndTests.js file inside of the static/test folder which holds the code that was previously in before() and after().

e2e-tests/home_e2e.js
cdoremus: Fixed

e2e-tests/home_e2e.js
cdoremus: Without calling done(), the test hangs. I'm not sure how to do what you are asking. Please advise.

Spacing error fixed.

package.json
cdoremus: Also changed the test name from home_e2e.js to home.js.

e2e-tests/home.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * End-to-end tests for home page
+ *
+ * See Readme.md for instructions on running the test
+ */
+const test = require('selenium-webdriver/testing');
+const chai = require('chai');
+const setup = require('../static/test/setupEndToEndTests');
```
Mr0grog: I kind of feel like code related to these tests should all live inside the `e2e-tests` directory. It’s pretty common to have a `e2e-tests/support/` directory that contains fixtures and setup code or library modules used by multiple tests. At the very least, I don’t think it really makes sense to mix code specific to e2e tests with code specific to unit tests.

(FWIW, the `test` directory is inside `static` instead of at the top level because the backend stuff was also in this repo at the time and I was trying to follow the directory structure that had all front-end-specific stuff in `static`. That placement does feel awkward now.)

e2e-tests/home.js
```diff
@@ -0,0 +1,37 @@
...
+test.describe('home page', () => {
+  let driver;
+
+  test.before(() => {
+    driver = setup.beforeTests();
+  });
+
+  test.after(() => {
+    setup.afterTests(driver);
+  });
```
Mr0grog: This feels a little verbose and repetitive. If you’re doing this to get at the `driver` object, there might be some nicer options:

- Have an accessor function that returns the current browser driver.
- Use the test context to store it, e.g:

  ```js
  test.before(function() { this.driver = ...; })
  ```

Either of the above would let you simplify these before/after calls to something like:

```js
test.describe('home page', function () {
  // this does the work of `test.before()` and `test.after()`
  setupBrowserTest();

  test.beforeEach(function () {
    getDriver().get(url);
    // or
    this.driver.get(url);
  });
  
  test.it('should show correct page title', function () {
    // test
  });
});
```

Or, as a decorator (kinda fancy):

```js
// `testWithBrowserDriver()` would be kind of like `setupBrowserTest()`
// from above when its return value is called
test.describe('home page', testWithBrowserDriver(() => {
  test.beforeEach(function () {
    getDriver().get(url);
    // or
    this.driver.get(url);
  });

  test.it('should show correct page title', function () {
    // test
  });
}));
```

You could also use Mocha’s `require` option to include setup code instead of using it like a library here (this is kind of what I thought you were going to do), but the approach you took here is actually nice in that it isolates a browser driver instance and its associated cache/storage/cookies to a given test suite (not sure if that will be too high overhead if/when running tests through a remote server like Sauce, though 🤔).

e2e-tests/home_e2e.js
Mr0grog: If you don’t plan to call `done` and want to return a promise instead, you need to not take `done` as an argument (Mocha checks the number of arguments your function takes and changes behavior based on that). So, either:

```js
test.it('should show correct page title', (done) => {
  let expected = '5 Calls: Make your voice heard';
  // no reason to return this; it doesn't accomplish anything
  driver.getTitle().then((title)=> {
    expect(title).to.equal(expected);
    done();
  });
});
```

OR, what I was suggesting:

```js
// note `done` is not an argument here
test.it('should show correct page title', () => {
  let expected = '5 Calls: Make your voice heard';
  // now you *must* return this promise, as its resolution or
  // rejection is what signals to mocha that the test is done
  return driver.getTitle().then((title)=> {
    expect(title).to.equal(expected);
  });
});
```

There’s also a nice Chai plugin, `chai-as-promised` that gives you some nicer promise-based assertions:

```js
// note again, `done` is not an argument here
test.it('should show correct page title', () => {
  let expected = '5 Calls: Make your voice heard';
  // the `eventually` bit is what chai-as-promised gives you
  // (note again you must return the result of this)
  return expect(driver.getTitle()).to.eventually.equal(expected);
});
```

package.json
cdoremus: The reason I (and others) like to use npm scripts is because I can be assured that everyone running the scripts are using the same version of gulp (the one in node_modules/.bin). This is particularly important for JS testing because chromedriver, karma and selenium can be very version sensitive.

Also, I don't know about you, but I do not update my global npm modules (like gulp) as often as I should.

e2e-tests/home.js
cdoremus: I moved the ```setupEndToEndTests.js``` file into the ```e2e-tests/support``` folder.

e2e-tests/home.js
cdoremus: I added the ```require``` option to the gulp test:e2e task pointing to ```setupEndToEndTests.js```. I also moved the ```test.before()``` and ```test.after()``` functions into this file.

package.json
Mr0grog: > I (and others) like to use npm scripts is because I can be assured that everyone running the scripts are using the same version of gulp (the one in node_modules/.bin).

That’s reasonable. I think there are better solutions to this problem (e.g. you should only install `gulp-cli` globally, which delegates to whatever gulp install is in your project), but like I said before, I don’t think it’s *bad* to have NPM pass-throughs. It’s just that there are 16 tasks, and if you really wanted to have npm scripts, you’d wind up having 16 `package.json` lines like:

```
"some-task": "gulp some-task"
```

…which seems kind of repetitive (and error prone) to me. I’m sure I’d make a typo that screws someone else over in at least one of those 16 lines.

> I don't know about you, but I do not update my global npm modules (like gulp) as often as I should.

I just avoid installing anything globally! 😛

Seriously, though, I have found the following in my shell startup script to be *super* useful:

```bash
# Node modules bin
NODE_PATH=./node_modules/.bin
for i in {1..5}
do
  path_level="../$path_level"
  NODE_PATH="$NODE_PATH:${path_level}node_modules/.bin"
done
export PATH="$NODE_PATH:$PATH"
```

…which means any time I run a command, my shell looks first to the closest `node_modules/.bin` directory (up to five levels up in the directory structure). I don’t expect anyone else to be doing that, but it really makes managing tooling dependencies much easier and more reliable.

e2e-tests/support/setupEndToEndTests.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Sets up mocha webdriver end-to-end tests.
+ *
+ * The beforeTests() function should be run inside a before() block
+ * with the return value assigned to a local variable.
+ *
+ * The afterTests() function should be run inside an after() block
+ * with its argument being the assigned value that was returned by
+ * beforeTests().
```
Mr0grog: These notes about `beforeTests()` and `afterTests()` are no longer relevant, right?

cdoremus: Yes. I'll make the change and push it to Git.

gulpfile.js
```diff
@@ -152,6 +153,14 @@ gulp.task('test:watch', function() {
   return runKarmaTests({singleRun: false});
 });
 
+gulp.task('test:e2e', function() {
+  return gulp.src(['./e2e-tests/**/*.js'])
```
Mr0grog: I don’t think we want to execute every file in `e2e-tests/support/` as if it was a test, right? I think this needs to be: `['./e2e-tests/**/*.js', '!./e2e-tests/support/**/*.js']`

gulpfile.js
```diff
@@ -152,6 +153,14 @@ gulp.task('test:watch', function() {
...
+    .pipe(mocha({
+      reporter: 'spec',
+      require: 'e2e-tests/support/setupEndToEndTests.js'
```
Mr0grog: This actually isn’t working. It needs to be `./e2e-tests/...` (note the `./` at the start so it doesn’t try to load an installed module) The file is only being run right now by virtue of it being counted as a test (see above).

That said, upon further looking into this, the method used here won’t work at all, so it was bad advice on my part. I’d only used this method to install globals and other helpers for use in tests, not to install setup/teardown handlers on Mocha. It turns out there’s no way to get a handle on the Mocha instance that is going to run the tests here, so it’s infeasible for that use. Sorry :(

Maybe better to include it as the first test file:

```js
return gulp.src([
  './e2e-tests/support/setupEndToEndTests.js',
  './e2e-tests/{*,!(support)/*}.js'
])
  .pipe(mocha({
    reporter: 'spec',
    timeout: 6000
  }));
```

Side note: I had to increase the `timeout` above because fontawesome’s CDN was frequently taking more than the standard 2000 ms allowed for each test. Not sure what’s best here, but so long as the tests will require actual remote resources, we probably need to up it at least a little.

Mr0grog:
👍

Mr0grog (approved):

cdoremus:
@Mr0grog thanks for your help. I learned a lot about configuring and running these tests in the process.

The next step is deciding how we are going to use these tests. My recommendation is to add them to the deploy pipeline rather than being part of every PR check. I suggest this because these tests can be slow and flaky. Still, I am not religious about this opinion, and I invite others to comment.

@nickoneill and @oderby what do you think?

BTW, I have put together a series of other tests for the Home page and added About and FAQ page tests that await merging this PR into master.

oderby:
Definitely want to hear from @nickoneill on this one, but I'm guessing the deploy pipeline is not yet automated. So while I'd be fine with pushing these tests to run as part of an automated deploy, I'd much rather see these run automatically and in a transparent place (so others can help debug), so they're reliably checking the code. Is there someplace downstream of PRs that these could be run automatically?

Also, @cdoremus have you already seen flakiness with these tests? Or are you just going off of previous experience, when suggesting these will be flaky?

nickoneill:
You're correct @oderby, there's no automated deployment set up. It's a single-command process at this point, "automated" in the makefile and I don't think there's additional benefit to be gained automating it beyond this at the moment.

I don't know what other part of the process this can be a part of but I could get behind it. Is there a reason it can't run on saucelabs? Do they not support selenium?

This new test run great for me currently. I'd be curious about the general flakiness as well but if it works well I'm happy to make it part of the PR process.

cdoremus:
@oderby and @nickoneill I started a new branch and put together 5 new tests (for item details, FAQ and About page) and they all run consistently, but it is my (and others) experience that e2e tests can be flaky. This has to do with the nature of these automated requests that are rapidly invoked much faster than a human client could ever do. Our tests pull from the production database, which adds latency to the responses from the automated requests. I have included wait calls in my new e2e tests that stops the test from proceeding until a condition is achieved (e.g. expected page elements or urls are rendered). These wait calls also have a timeout, which I have set to 2 or 3 seconds. They help to minimize flakiness. Still, you could set all the waits to 30 seconds and still have a failing test because the browser is hanging due to client computer, server or network issues.

Adding to the uncertainty is the dynamic nature of the data being sent back by the production server. Ideally, these kind of tests should be run using a dedicated test server and and a consistent test data set.

Despite these caveats, I certainly can sign on to adding these tests as a PR check now. We can always modify the tests or how they fit into the development/deployment pipeline later.

oderby:
@nickoneill As I understand it ( @Mr0grog and @cdoremus please correct me if I'm wrong) Sauce Labs integration will come next. This first task was just about getting the system set up and configured locally for the project. 

But, that brings up a good point - these can't be run as part of the PR process (or any automated process) until we get Sauce Labs/CI integration set up, right?

Mr0grog:
> As I understand it ( @Mr0grog and @cdoremus please correct me if I'm wrong) Sauce Labs integration will come next.

That is definitely what I thought the plan was.

> these can't be run as part of the PR process (or any automated process) until we get Sauce Labs/CI integration set up, right?

They can—it’s totally possible to run a browser inside Circle—in fact, I think it has Chrome already available, though it’s not always very up-to-date (most projects have a script to install the latest Chrome as part of their pre-test setup).

I do think these should be PR checks. The whole point of having a massive and reliable CI service is to be able to handle heavy/long builds and tests that are infeasible on your local machine. Even if they’re slow, you still want to run all your tests before merging. You may as well have the CI server do them.

> Ideally, these kind of tests should be run using a dedicated test server and and a consistent test data set.

Totally. Also with local versions of assets that might otherwise come from a CDN (e.g. fontawesome). This PR had gotten into a pretty long back-and-forth, so I was thinking that would probably be better as a second PR.

Mr0grog:
re: the browser/Circle/Sauce thing: you still ultimately want sauce because you want to test in multiple versions of browsers and across multiple platforms; you only have the OS you’re running on when running a browser directly on Circle. Doing that can be a quick intermediary step to getting Sauce going right.

nickoneill:
OK, I think we're all on the same page. I'm working on getting a test/staging server up and running, not sure about a consistent data set but at least we'll have some way to deploy new stuff to test without going straight to prod.

Are we good to merge this now or did we need an additional change for this to run by default for PRs? @cdoremus or @Mr0grog, can one of your summarize testing next steps in an issue so we can track going forward.

cdoremus:
@nickoneill I think this PR is ready to merge. Since the single test in the PR is somewhat of a 'hello world' example, I have some other tests that I've worked on that better exercise an e2e test system. All of these tests can be run manually, and I have included instructions in the ```Readme``` in this PR on how to do that. At the same time we need Sauce Labs and CircleCI configured, which is best done by Rob (@Mr0grog).

Once this PR goes in, I'll close out #207 and open a new one for further discussions. I can also create a new PR containing my new e2e tests if that is agreeable to the group.

nickoneill:
Perfect, thanks @cdoremus!

PR #211, opened 2017-02-22
Fix #197 - add support for rendering call count on Safari

### LINKED ISSUE #197: Support Safari 9
Looks like Safari 9 is missing the `toLocaleString` feature we use to format our numbers nicely. It has roughly 50k views in the last month so it would be nice to support this browser version. Tests are already in place, just need to figure out if there's a polyfill or something we can use here.

oderby:
Given that we have a restricted subset of the toLocaleString
functionality we need (just inserting thousands separator) and
we only support a single locale now ('en') and that we are a
US-centric app, I decided to just special case handling in
the component for now.

### DIFF
--- a/static/js/components/callcount.js
+++ b/static/js/components/callcount.js
@@ -8,9 +8,12 @@ module.exports = (state, prev, send) => {
   `;
 
   function callCount(state) {
-    const calls = Number(state.totalCalls);
+    let calls = Number(state.totalCalls);
+    // Handle undefined input.
+    // Number(undefined) is NaN, while Number("") is 0
+    calls = isNaN(calls) ? 0 : calls;
     // Number.toLocaleString() doesn't work on Safari 9 (see https://github.com/5calls/5calls/issues/197)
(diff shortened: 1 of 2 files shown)


nickoneill (approved):
👍

nickoneill:
Will bring this in for the next frontend deploy

cdoremus (changes requested):
It would be good to have unit test coverage on the callCount() function, especially the regex.
Also, why the double negative (!!) on the window.Intl expression in the if statement? It's confusing.

oderby:
@cdoremus there already are tests, in callcount_test.js. What's missing that you'd like to see? The only other one that comes to mind is malformed/missing state/dependencies (that is when `state.totalCalls` is undefined). I'll add that, unless you have something additional in mind?

regarding `!!` you're right, not needed in this case. I'll remove.

cdoremus (approved):
Tests are OK. Thanks for removing the double negative.

PR #214, opened 2017-02-23
Handle local offices without cities

nickoneill:


oderby (changes requested):
Test?

nickoneill:
Added a test, my first shot at chai, be gentle @oderby

static/js/components/contact_test.js
```diff
@@ -81,4 +81,25 @@ describe('contact component', () => {
...
+      photoURL: '',
+      reason: '',
+      field_offices: field_offices
+    };
+    let state = {
+      showFieldOfficeNumbers: true
+    };
+    let result = contact(contactData, state);
+    let content = result.querySelectorAll('ul.call__contact__field-office-list li a')[0].textContent;
+    expect(content).to.equal(field_offices[0].phone);
```
oderby: This is only testing that when field office data is missing the city, the phone number is still present. You probably also want to check that the state abbreviation is NOT included. Like so:

```javascript
    let content = result.querySelectorAll('ul.call__contact__field-office-list li')[0].textContent;
    expect(content).to.contain(field_offices[0].phone);
    expect(content).to.not.contain(contactData.state);
```

(Note I changed the query selector to not grab just the `a` element, so we can be sure we're checking the entire list item.)

static/js/components/contact_test.js
```diff
@@ -81,4 +81,25 @@ describe('contact component', () => {
     expect(content).to.contain(field_offices[0].phone);
     expect(content).to.contain(field_offices[0].city);
```
oderby, line 82: We should check that the state is present here.

```javascript
    expect(content).to.contain(contactData.state);
```

static/js/components/contact.js
```diff
@@ -23,14 +23,22 @@ module.exports = (c, state, prev, send) => {
...
+              <li><a href="tel:+1${office.phone.replace(/-/g, '')}">${office.phone}</a> ${cityFormat(office,c)}</li>
             `)}
           </ul>
         </div>
       `
     }
   }
 
+  function cityFormat(office, c) {
+    if (office.city != "") {
```
oderby: no need to compare to empty string. javascript will coerce to boolean based on truthiness of the value, and "" is falsey. So just do `if (office.city)`. This way, you'll also have proper behavior if `office.city===undefined` or `office.city===null`, since those both are falsey. Isn't javascript wonderful??

oderby (changes requested):
small fix to tests, and nit on (ab)using javascript boolean coercion. Otherwise lgtm!

nickoneill:
OK! Thanks for the tips, I was wondering about doing some alternate tests and this clears things up.

oderby (approved):

Issue #216, opened 2017-02-23
Pressing back button breaks the app


oderby:
# Repro Steps
1. Go to [5calls.org](5calls.org)
2. click on first issue (should get to url such as `https://5calls.org/#issue/recmbgLuYnXWoJhQq`)
3. click browser back button
# Result
URL is updated to malformed url (e.g. `https://5calls.org/#issuerecmbgLuYnXWoJhQq"`) and get a stack trace in the console:
![image](https://cloud.githubusercontent.com/assets/1351158/23252661/7d5aa558-f966-11e6-8e76-e0c7b4827ca5.png)
Clicking on any further links has no effect. 
Reloading page also results in same error and broken app state.
# Expected
It should navigate back to home page (or whatever page user was previously on)
# Environment
Chrome 55 and Firefox 51
# Notes
This is really two bugs
1. Broken history
2. unmatched routes throw an uncaught error, crashing the entire app. This may be something we should report upstream to [choo](https://github.com/yoshuawuyts/choo), but should definitely see if we can't handle locally.

nickoneill:
FYI this previously worked OK, not sure what broke it but it's very particular.

sshevlyagin:
Tried to repro. Looks like it's no longer an issue.

nickoneill:
Yep, all set here. Thanks for the reminder @sshevlyagin!

PR #219, opened 2017-02-24
typo - missing quote and slash on href value.

beausmith:
Introduced here: https://github.com/5calls/5calls/commit/6e4341390876c2936ac528bfd2b560655a9292db#diff-13f2f55f8c5ad0a842db049e9f8b0caaR27

cc @liamdanger

### DIFF
--- a/static/js/components/issuesListItem.js
+++ b/static/js/components/issuesListItem.js
@@ -36,7 +36,7 @@ module.exports = (issue, state, prev, send) => {
 
   return html`
     <li onclick=${handleClick}>
-      <a aria-controls="content" class="${classString(state, '')}" href=#issue${issue.id}">
+      <a aria-controls="content" class="${classString(state, '')}" href="#issue/${issue.id}">
         <span aria-live="polite" class="${classString(state, '__status')}"><span class="visually-hidden">${statusText}</span></span>
         <span class="${classString(state, '__title')}">${issue.name}</span>
         <span class="${classString(state, '__summary')}">${issue.contacts.length} call${ issue.contacts.length > 1 ? "s" : "" } to make</span>


oderby:
This is definitely related to #216 - the url I see when I go back is exactly of form `issue{issueid}"`... Will pull down and test in a second.

nickoneill:
Ah, awesome. I looked around for this the other night in the JS functions for changing location, didn't even think to check the html itself.

oderby (approved):
lgtm

oderby:
So this definitely alleviates the worst part of #216, but it still takes two back clicks to go back to the previous issue/page. So let's leave #216 open and use it to track fixing the rest of the issue.

nickoneill:
OK, this solution got me thinking. The double-back issue is because we have a handleClick AND an anchor that are doing the same thing. So effectively we're following the same link twice every time we click on an issue and choo adds the browser history twice.

Now that we have testing, we can pick one or the other and test across browsers so make sure we're not breaking IE 10 (which I think might be the reason that anchor got there in the first place).

oderby:
@nickoneill You're right that we only need one or the other. It seems to me the best option is to leave the anchor and just remove [main.js:L377](https://github.com/5calls/5calls/blob/master/static/js/main.js#L377). 

But I don't follow what you think we should be testing on IE10?

nickoneill:
Just tapping on the issue and getting the detail screen to show up. IIRC, some versions of IE don't like onclicks on non-anchor elements? If the tests run OK I'm good, I'm literally pulling at information I learned about IE more than 6 years ago to hypothesize here.

oderby:
Okay, will need to add that to the growing list of e2e tests to write once we get #209 up and running. For now, I'll push a fix for the double-back issue.

Issue #213, opened 2017-02-23
Add caching headers to issue responses


nickoneill:
Can be based on contacts and issue generation

PR #220, opened 2017-02-25
Users can view issues not featured in the main sidebar

beausmith:
@oderby - Please review this PR. The only thing I didn't get to is renaming various "issues" files to "activeIssues" files. 

**Note** - This issue requires new endpoint that @oderby working on.

### DIFF
--- a/static/js/components/call.js
+++ b/static/js/components/call.js
@@ -9,9 +9,11 @@ const promote = require('./promote.js');
 
 module.exports = (state, prev, send) => {
   const issue = find(state.issues, ['id', state.location.params.issueid]);
-
   if (issue == null) {
-    return html`<section class="call" onload=${(e) => send('oldcall')}>
+    return html`<section class="call" onload=${(e) => {
+      send('fetchInactiveIssues')
+      send('oldcall')
+    }}>
       <div class="call_complete">
         <h2 class="call__title">No calls to make</h2>
         <p class="call__text">
--- a/static/js/components/issues.js
+++ b/static/js/components/issues.js
@@ -1,4 +1,5 @@
 const html = require('choo/html');
+const scrollIntoView = require('scroll-into-view');
 
 const issuesHeader = require('./issuesHeader.js');
 const issuesList = require('./issuesList.js');
@@ -12,10 +13,15 @@ module.exports = (state, prev, send) => {
     send('resetCompletedIssues');
   }
 
+  function scrollToTop () {
+    scrollIntoView(document.querySelector('#content'))
+  }
+
   return html`
     <div class="issues">
       ${issuesHeader(state, prev, send)}
       ${issuesList(state, prev, send)}
+      <a href="#issues" class="issues__footer-link" onclick=${scrollToTop}>view more issues</a>
       ${debugText(state.debug)}
     </div>
   `;
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -6,6 +6,7 @@ const logger = require('loglevel');
 const queryString = require('query-string');
 const store = require('./utils/localstorage.js');
 const scrollIntoView = require('scroll-into-view');
+const issueCacheMinutes = 30;
 
 const app = choo();
 const appURL = 'https://5calls.org';
@@ -79,6 +80,10 @@ app.model({
   state: {
     // remote data
     issues: [],
+    activeIssues: [],
+    inactiveIssues: [],
+    lastActiveIssueFetchDate: null,
+    lastInactiveIssueFetchDate: null,
     totalCalls: 0,
     splitDistrict: false,
 
@@ -107,14 +112,30 @@ app.model({
   },
 
   reducers: {
-    receiveIssues: (state, data) => {
+    receiveActiveIssues: (state, data) => {
       response = JSON.parse(data)
-      issues = response.issues //.filter((v) => { return v.contacts.length > 0 });
-      contactIndices = {};
-      issues.forEach(function(item, index) {
-         contactIndices[item.id] = 0;
-      });
-      return { issues: issues, splitDistrict: response.splitDistrict, invalidAddress: response.invalidAddress, contactIndices: contactIndices }
+      return {
+        activeIssues: response.issues,
+        splitDistrict: response.splitDistrict,
+        invalidAddress: response.invalidAddress,
+      }
+    },
+    receiveInactiveIssues: (state, data) => {
+      response = JSON.parse(data)
+      return {
+        inactiveIssues: response.issues,
+      }
+    },
+    mergeIssues: (state, data) => {
+      issues = state.activeIssues.concat(state.inactiveIssues)
+      contactIndices = state.contactIndices
+      issues.forEach(issue => {
+        contactIndices[issue.id] = 0
+      })
+      return {
+        issues,
+        contactIndices,
+      }
     },
     receiveTotals: (state, data) => {
       totals = JSON.parse(data);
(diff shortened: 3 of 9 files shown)


nickoneill:
I checked this with @oderby's backend changes, it's an excellent start. I can get the issue details to show up when directly linking but the list of issues both on the main / more issues page don't show anything. It looks like the issues are loading correctly (during the merge phase there are items in active and inactive lists) but not displayed? No console errors.

Some other things I noticed:
* location spins forever on the all issues page
* it looks like you added issue caching in here too but I didn't see a spot where it's invalidated if the location changes (i.e. new contacts, which are attached directly to issues as of now)

static/js/main.js
```diff
@@ -105,14 +110,30 @@ app.model({
...
+    receiveInactiveIssues: (state, data) => {
+      response = JSON.parse(data)
+      return {
+        inactiveIssues: response.issues,
+      }
+    },
+    mergeIssues: (state, data) => {
+      issues = state.activeIssues.concat(state.inactiveIssues)
+      contactIndices = state.contactIndices
+      issues.map(issue => {
```
oderby: You should use `forEach` instead of `map` here, since this is purely side-effects.

static/js/main.js
```diff
@@ -105,14 +110,30 @@ app.model({
...
+        contactIndices[issue.id] = contactIndices[issue.id] || 0
```
oderby: Why the change to keep old contactIndices? If I understand it, the previous code would always reset the contact indices. Now we won't. This change means that if someone changes their location after making it part way through an issue, they won't see the new contacts. There's probably some other complexities here, but unless there's a compelling reason to, I don't think we should change this behavior in this PR.

nickoneill:
🏓

static/js/main.js
beausmith: Cool.

static/js/main.js
beausmith: Reverted. I thought this was for another purpose, now I see what it's really for. Reverting.

beausmith:
Did a quick look into the issues @nickoneill reported, was able to repro, but wasn't able to fix. Will look into it again tonight/tomorrow. If someone looks into it sooner, it's all good.

nickoneill:
Fixed this issue on the saucelabs side, next time it runs should be OK.

static/js/main.js
```diff
@@ -188,23 +209,48 @@ app.model({
...
   effects: {
-    fetch: (state, data, send, done) => {
-      address = "?address="
-      if (state.address !== '') {
-        address += state.address
-      } else if (state.geolocation !== "") {
-        address += state.geolocation
+    fetchActiveIssues: (state, data, send, done) => {
+      if (((Date.now() - Number(state.lastActiveIssueFetchDate)) / 1000 / 60) < issueCacheMinutes && !!state.activeIssues.length) {
+        return
```
oderby: So looking at this with fresh eyes, what benefits do we get from manually caching issues on the client? With #213, the browser should cache the request for us automatically, and the merge logic is trivial and should run more than fast enough. Until that lands, it'd mean we'll fetch inactive issues the first time they load an inactive issue and each time thereafter they load the `/issues` page. I wonder if a slow-loading `/issues/` page is worth the reduction in code complexity? 

We could even go further and only fetch (in)active issues once, (by checking here and in `fetchInactiveIssues` if any (in)active issues are present in `state.issues`). This would mean they wouldn't see newly-added or changed (active<->inactive) issues without refreshing the page, but that seems like a really low-frequency event.

beausmith:
Ok! @nickoneill I have fixed the issue you reported regarding loading the "more issues" page manually. I've also updated all tests to pass. Yay tests!

static/js/main.js
beausmith: @oderby - I feel ya. 

Because #213 isn't done… should we make a note there to remove this manual caching? …or should I remove it now in anticipation of that task being completed?

I had thought about using the existence of (in)active issues as the cache control, but then thought that this solution would lead to less end-user confusion. 

Ok, I have removed the caching in a separate commit (https://github.com/5calls/5calls/pull/220/commits/ce7271d9b12108aac5e70ddfa27273b96986a574) in the case that someone else wants it back. Either way, I think this PR should be ready to roll.

static/js/components/call.js
```diff
@@ -9,9 +9,11 @@ const promote = require('./promote.js');
 
 module.exports = (state, prev, send) => {
   const issue = find(state.issues, ['id', state.location.params.issueid]);
-
   if (issue == null) {
-    return html`<section class="call" onload=${(e) => send('oldcall')}>
+    return html`<section class="call" onload=${(e) => {
+      send('fetchInactiveIssues')
```
oderby, line 14: This feels a little hacky, but it's not incorrect and, for most people, will only result in a quick flash of "bad issue" text before we load the inactive issues list and display the issue content. If we start seeing people on slow connections or this becomes a problem, we can add an intermediate text saying "loading your issue, hold tight" or such.

static/js/components/issues.js
```diff
@@ -12,10 +13,15 @@ module.exports = (state, prev, send) => {
     send('resetCompletedIssues');
   }
 
+  function scrollToTop () {
+    scrollIntoView(document.querySelector('#content'))
```
oderby, line 17: Will this need to change with #226?

static/js/main.js
```diff
@@ -321,7 +354,7 @@ app.model({
         }
```
oderby, line 354: Line 346 above should probably just look at `state.activeIssues.length`, right? I don't think it matters in practice, but there's technically a race between that check and the `fetchInactiveIssues` action and ensuing `mergeIssues` actions completing.

oderby (commented):
This is looking good. Just one or two small changes, otherwise I didn't find any bugs in my limited testing. @nickoneill we shouldn't actually merge this until the corresponding backend changes are merged and deployed. Think we can go ahead and start pushing that through while this goes through final iterations?

static/js/components/issues.js
bernardwang: It shouldn't! The function scrollIntoView has been redefined, but it is used exactly the same.

static/js/components/call.js
nickoneill: Same thing occurs today when linking to an issue, so I don't think this is a huge deal.

static/js/main.js
nickoneill: I'm looking at line numbers and I'm not sure I understand where you're referencing here.

nickoneill (approved):

static/js/main.js
oderby: yeah, it sucks that github doesn't let you comment on lines not within the original diff context... I'm referring to:
```javascript
       if (state.issues.length == 0 || state.geolocation == '') {
```
and suggesting it should be changed in this PR to read
```javascript
       if (state.activeIssues.length == 0 || state.geolocation == '') {
```

static/js/components/call.js
oderby: :+1:

nickoneill:
Looks great to me! I'd like to make sure @oderby is good on that last comment he had but otherwise ready to go out.

Seeing the list was making Rebecca think about some more organization on that page, particularly if it starts to have too many inactive issues. Maybe some categories in Airtable are in order (in the future).

beausmith:
@oderby & @nickoneill - I made that last change which @oderby pointed out. 

Looking forward to iterating on this page more once we get this basic functionality out there.

nickoneill:
Thank you @beausmith and @oderby!

PR #221, opened 2017-02-25
Remove duplicate navigation when selecting issue.

oderby:
This fixes an issue where, after selecting an issue,
it would take two presses of the browser back button
to navigate back to the previous page. This issue was
uncovered (indirectly) while triaging #216.

Improved unit test coverage, but, as per discussion
in #219, we should add an e2e test to verify that
clicking on the link successfully navigates to the
issue.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -6,6 +6,7 @@ const logger = require('loglevel');
 const queryString = require('query-string');
 const store = require('./utils/localstorage.js');
 const scrollIntoView = require('scroll-into-view');
+const onHashChange = require('sheet-router/hash');
 
 const app = choo();
 const appURL = 'https://5calls.org';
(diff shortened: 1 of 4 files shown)


oderby:
Okay, updated with the most correct fix. This is a gnarly bug at the intersection of using hash routing with choo on IE. I've tested this PR on both chrome and IE. Would appreciate someone to review by testing on safari and making sure issue navigation works on safari as well (including fowards/backwards navs work properly). I'

static/js/main.js
```diff
@@ -370,15 +371,27 @@ app.model({
...
+    // Problem: Choo's internal handlers for updating the internal location model
+    // do not work properly with hash routes on IE, due to two separate bugs:
+    //  1. popstate event is not fired when only the hash changes (https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/3740423/)
+    //     meaning the internal history handler doesn't fire.
+    //  2. the href handling logic has a bug (https://github.com/yoshuawuyts/sheet-router/issues/82)
+    //     so the internal href handler also doesn't fire
+    // The result of these two issues is that when a user clicks on an issue
+    // in our app on IE, the URL in the browser changes, but the app doesn't
+    // update to render the new view.
+    // Solution: We explicitly esnure choo's internal location model is correct
```
oderby: "ensure"

static/js/main.js
```diff
@@ -6,6 +6,7 @@ const logger = require('loglevel');
 const queryString = require('query-string');
 const store = require('./utils/localstorage.js');
 const scrollIntoView = require('scroll-into-view');
+const onHashChange = require('sheet-router/hash');
```
oderby, line 9: Do we need to add 'sheet-router' to package.json?

oderby (commented):
Some small comments for myself. I also think we should hold off on this until #225 lands. The hash change subscription will still be useful, since we'll still have other hash routes (/#about, /#issues) that will need the support.

static/js/main.js
nickoneill: Dependency of choo, I believe.

nickoneill:
The change subscription feels hacky to me, I'd prefer to swap the urls to non-hash on the backend if we have to keep it otherwise.

oderby:
I'm just copying how choo itself does it, so it's as hacky as choo is... But agreed, if we can just remove all hash routing, then all the better.

static/js/main.js
oderby: Yeah, but I'm pretty sure best practice is to explicitly list all your direct dependencies. This adds sheet-router as such, and so it should be added to the `package.json`

oderby:
Fixed in #256

PR #225, opened 2017-02-26
Remove hash urls to issues

nickoneill:
With associated backend changes

### DIFF
--- a/static/js/components/issuesListItem.js
+++ b/static/js/components/issuesListItem.js
@@ -36,7 +36,7 @@ module.exports = (issue, state, prev, send) => {
 
   return html`
     <li onclick=${handleClick}>
-      <a aria-controls="content" class="${classString(state, '')}" href="#issue/${issue.id}">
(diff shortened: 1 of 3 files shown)


static/js/components/issuesListItem.js
```diff
@@ -36,7 +36,7 @@ module.exports = (issue, state, prev, send) => {
 
   return html`
     <li onclick=${handleClick}>
-      <a aria-controls="content" class="${classString(state, '')}" href="#issue/${issue.id}">
```
oderby, line 39: As discussed online, should keep the href and remove the 'location:set' action from main.js instead.

oderby:
@nickoneill ping - what's blocking here? How can I help? With the release of the "more issues" page, the back button problem is more apparent on the site, so I'd like to get #221 in, in some form. If this is blocked, I say we merge #221 as is (I've just tested it rebased onto current master and it still works correctly) and then remove the onHashChange code here. Thoughts?

nickoneill:
Not blocked, just timing. Working on this now.

nickoneill:
remove in favor of #256

Issue #208, opened 2017-02-19
Make scrolling smoother


bernardwang:
Two ways to do this:
1.  Cancelling the autoscroll when scrolling manually
2.  Check if target is in viewpoint before scrolling

PR #226, opened 2017-02-27
Smoother scrolling

bernardwang:
## Summary
Fixes issue [208](https://github.com/5calls/5calls/issues/208)

1. The page will only scroll if the element is not at the top of the viewport.
2. Scrolling is smoother in general, especially in Firefox. It is still slightly jumpy in Chrome.
3. The transition speed is decreased from 1000ms to 500ms (due to the defaults of the scrolling dependencies).

Replaces [`scroll-into-view`](https://www.npmjs.com/package/scroll-into-view) with [`smoothscroll-polyfill`](https://www.npmjs.com/package/smoothscroll-polyfill). The two dependencies both use [`raf`](https://www.npmjs.com/package/raf) and are implemented similarly. However using a polyfill has some slight advantages like being natively supported in Firefox.

For reasons I haven't yet figured out, the ```element.scrollIntoView``` function is still jumpy in Chrome. The issue doesn't happen in Safari, Firefox, or with ```window.scroll```. Because of this I created a util file ```scrollIntoView.js```, which calculates the scroll position of an element, and then uses ```window.scroll```.

## Testing
Videos of scrolling:
[Chrome](https://i.gyazo.com/45cc692eb2a395d985b577689456a767.mp4)
[Safari (responsive view)](https://i.gyazo.com/10f9a4f554b16aa1d8caa5f6ca349e3c.mp4)
[Firefox](https://i.gyazo.com/2771867d425f3b6f3d709d07943aab82.mp4)


nickoneill:
Thanks @bernardwang! I'll check this in the next day or so. It doesn't look like that test failure was related, do you get the same thing when running tests locally?

bernardwang:
I pulled and tests are now OK!

static/js/utils/scrollIntoView.js
```diff
@@ -0,0 +1,12 @@
+
+const raf = require('raf').polyfill();
+const smoothscroll = require('smoothscroll-polyfill').polyfill();
```
oderby: Is there any benefit to capturing the `undefined` both `polyfill` methods return? In other words, why shouldn't this just be

```javascript
// load polyfills for window.requestAnimationFrame and window.scroll as side-effects.
require('raf').polyfill();
require('smoothscroll-polyfill').polyfill();
```

oderby:
I still want to test it out on my IE when I get a chance later today, but looks fine to me.

oderby:
It looks fine on IE.

Would still like to see my comment addressed or answered, but beyond that, this seems fine to me.

bernardwang:
OK thanks @oderby! Which comment are you referring to? I only saw the comment on the unnecessary return saving.

oderby (approved):

oderby:
All set - great work @bernardwang !

PR #231, opened 2017-03-04
small fix to app icons in IE 11

oderby:
Currently, when you load 5calls.org on IE 11, you see the IOS app icon is stretched and misaligned:
image

This seems like a bug in IE to me, and this change works around the issue. However, I'll admit I'm not really a CSS guru, so I'm not sure what exactly is wrong or if this is the best fix.

### DIFF
--- a/static/scss/_hypothesis.scss
+++ b/static/scss/_hypothesis.scss
@@ -49,11 +49,11 @@
     }
 
     .ios {
-      padding: 12px 0 0 0;
-      width: 174px;
+      padding: 9px 0 0 0;
+      width: auto;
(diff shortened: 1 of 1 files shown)


nickoneill:
Thanks! The bliss of not having an IE 11 machine anywhere in my house is coming back to bite me.

Test failure looks like a saucelabs issue, I just restarted it.

nickoneill (approved):

nickoneill:
Verified on IE11 via browserstack, looks fine in Firefox, chrome and safari on my machine as well. Thanks!

PR #232, opened 2017-03-05
Improved some sharing behavior

nickoneill:
- Added sharing buttons to the front page with default sharing text
- Both share links should pop a new window for sharing
- Tests for the differences between issue / generic shares

### DIFF
--- a/static/js/components/promote.js
+++ b/static/js/components/promote.js
@@ -16,11 +16,21 @@ module.exports = (state, prev, send, issue) => {
     facebookTitle = "Share this issue"
   }
 
+  function tweetShare(e) {
+    e.preventDefault();
+    window.open("https://twitter.com/share?url="+url+additionalComps+"&text="+tweet, 'sharewindow', 'width=500, height=350');
+  }
+
+  function fbShare(e) {
+    e.preventDefault();
+    window.open("https://www.facebook.com/sharer/sharer.php?u=http://bit.ly/2iJb5nH", 'sharewindow', 'width=500, height=350');
+  }
+
   return html`
     <div class="promote">
       <p>
-        <a target="_blank"
-          href="https://twitter.com/share?url=${url}${additionalComps}&text=${tweet}"><i class="fa fa-twitter" aria-hidden="true"></i> ${twitterTitle}</a> <a target="_blank" href="https://www.facebook.com/sharer/sharer.php?u=http://bit.ly/2iJb5nH"><i class="fa fa-facebook" aria-hidden="true"></i> ${facebookTitle}</a>
+        <a target="_blank" onclick=${(e) => tweetShare(e)}><i class="fa fa-twitter" aria-hidden="true"></i> ${twitterTitle}</a>
+        <a target="_blank" onclick=${(e) => fbShare(e)}><i class="fa fa-facebook" aria-hidden="true"></i> ${facebookTitle}</a>
       </p>
     </div>
   `;  
--- a/static/js/components/promote.js
+++ b/static/js/components/promote.js
@@ -1,9 +1,9 @@
 const html = require('choo/html');
 
 module.exports = (state, prev, send, issue) => {
-  let url = "https://5calls.org?utm_campaign=twshare"
-  let additionalComps = "&via=make5calls"
-  let tweet = "try a make5calls today"
+  let url = encodeURIComponent("https://5calls.org?utm_campaign=twshare")
+  let additionalComps = encodeURIComponent("&via=make5calls")
+  let tweet = encodeURIComponent("try a make5calls today")
   let twitterTitle = "Share on Twitter"
   let facebookTitle = "Share on Facebook"
 
--- a/static/js/components/promote.js
+++ b/static/js/components/promote.js
@@ -1,14 +1,22 @@
 const html = require('choo/html');
+const find = require('lodash/find');
 
-module.exports = (state, prev, send, issue) => {
-  let url = encodeURIComponent("https://5calls.org?utm_campaign=twshare")
-  let additionalComps = encodeURIComponent("&via=make5calls")
+module.exports = (state, prev, send) => {
+  // share widgets for both platforms don't play well with dom manipulation so we'll make our own
+
+  // if no issue is selected, use these default parameters
+  let url = encodeURIComponent("https://5calls.org")
+  let additionalTwitterComps = "&via=make5calls"
   let tweet = encodeURIComponent("Spend 5 minutes. Make 5 calls. Make your voice heard.")
   let twitterTitle = "Share on Twitter"
   let facebookTitle = "Share on Facebook"
 
+  const issue = find(state.issues, ['id', state.location.params.issueid]);
+
+  // for selected issues, customize the share text a bit more
   if (issue) {
-    url = encodeURIComponent('http://5calls.org/#issue/' + issue.id + '?utm_campaign=twshare')
+    url = encodeURIComponent('http://5calls.org/#issue/' + issue.id)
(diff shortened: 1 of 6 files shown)


static/js/components/promote.js
```diff
@@ -1,17 +1,37 @@
 const html = require('choo/html');
 
 module.exports = (state, prev, send, issue) => {
+  let url = encodeURIComponent("https://5calls.org?utm_campaign=twshare")
```
oderby: I thought we [had decided](https://github.com/5calls/5calls/pull/150#discussion_r100710474) to stick with keeping component signatures as `(state, prev, send)` and looking up the current issue from within whichever components needed it.

static/js/components/promote.js
```diff
@@ -1,17 +1,37 @@
...
+  let twitterTitle = "Share on Twitter"
+  let facebookTitle = "Share on Facebook"
+
   if (issue) {
-    const url = encodeURIComponent('http://5calls.org/#issue/' + issue.id + '?utm_campaign=twshare');
-    const tweet = encodeURIComponent('I just called my rep to ' + issue.name.substring(0, 72) +
-    ' — you should too:');
-    return html`
+    url = encodeURIComponent('http://5calls.org/#issue/' + issue.id + '?utm_campaign=twshare')
+    additionalComps = ""
```
oderby: It's not obvious to me why we want to include our twitter handle for generic shares but not issue-specific ones. Perhaps worth a comment?

static/js/components/promote.js
```diff
@@ -1,17 +1,37 @@
 const html = require('choo/html');
 
 module.exports = (state, prev, send, issue) => {
+  let url = encodeURIComponent("https://5calls.org?utm_campaign=twshare")
+  let additionalComps = encodeURIComponent("&via=make5calls")
```
oderby: as far as I can tell, this is only for twitter-specific arguments, so perhaps rename to `twitterAddtionalArgs`?

static/js/components/promote.js
```diff
@@ -1,17 +1,37 @@
...
+  let tweet = encodeURIComponent("Spend 5 minutes. Make 5 calls. Make your voice heard.")
+  let twitterTitle = "Share on Twitter"
+  let facebookTitle = "Share on Facebook"
+
   if (issue) {
-    const url = encodeURIComponent('http://5calls.org/#issue/' + issue.id + '?utm_campaign=twshare');
```
oderby: What is the twshare keyword used for? Perhaps worth a comment (especially since it suggests it's only used for twitter, assuming it's short for "twitter share", but we use it for both twitter and facebook shares)

static/js/components/promote.js
```diff
@@ -1,17 +1,37 @@
...
+    additionalComps = ""
+    tweet = encodeURIComponent('I just called my rep to ' + issue.name.substring(0, 72) +
+    ' — you should too:')
+    twitterTitle = "Tweet this issue"
+    facebookTitle = "Share this issue"
+  }
+
+  function tweetShare(e) {
+    e.preventDefault();
+    window.open("https://twitter.com/share?url="+url+additionalComps+"&text="+tweet, 'sharewindow', 'width=500, height=350');
```
oderby: What's the rationale for opening a new window vs. what we have now (opening in a new tab)? General guidance and usability [strongly discourages](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Avoid_resorting_to_window.open()) opening new windows.

static/js/components/promote.js
nickoneill: Yeah, we definitely did. I don't think I actually modified it here but I'll refactor to do the right thing.

oderby: my "you touched it" monitor has a fuzzy filter of +/- 5 lines ;)

static/js/components/promote.js
nickoneill: I actually swapped in the official sharing widgets for both platforms for a while here but they don't play nice with choo (rather, constantly rebuilding the DOM) and this seemed most consistent with that behavior. I generally agree with this advice for opening new windows/tabs but post creation pages on both of these sites look awful when opened in full tabs.

oderby:
Makes sense. Just add a summary of that logic in a comment?

On Mar 5, 2017 5:12 PM, "nickoneill" <notifications@github.com> wrote:

*@nickoneill* commented on this pull request.
------------------------------

In static/js/components/promote.js
<https://github.com/5calls/5calls/pull/232#discussion_r104331466>:

>    if (issue) {
-    const url = encodeURIComponent('http://5calls.org/#issue/' +
issue.id + '?utm_campaign=twshare');
-    const tweet = encodeURIComponent('I just called my rep to ' +
issue.name.substring(0, 72) +
-    ' — you should too:');
-    return html`
+    url = encodeURIComponent('http://5calls.org/#issue/' + issue.id +
'?utm_campaign=twshare')
+    additionalComps = ""
+    tweet = encodeURIComponent('I just called my rep to ' +
issue.name.substring(0, 72) +
+    ' — you should too:')
+    twitterTitle = "Tweet this issue"
+    facebookTitle = "Share this issue"
+  }
+
+  function tweetShare(e) {
+    e.preventDefault();
+    window.open("https://twitter.com/share?url="+url+additionalComps+"&text="+tweet,
'sharewindow', 'width=500, height=350');

I actually swapped in the official sharing widgets for both platforms for a
while here but they don't play nice with choo (rather, constantly
rebuilding the DOM) and this seemed most consistent with that behavior. I
generally agree with this advice for opening new windows/tabs but post
creation pages on both of these sites look awful when opened in full tabs.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/5calls/5calls/pull/232#discussion_r104331466>, or mute
the thread

nickoneill:
All comments attended to, except for twshare which I realized is unnecessary so it's been removed.

static/js/components/promote_test.js
```diff
@@ -4,18 +4,19 @@ const chai = require('chai');
 const expect = chai.expect;
 
 describe('promote component', () => {
-  it('should NOT display tweet link if issue does not exist', () => {
+  it('should display generic tweet link if issue does not exist', () => {
     let issue = undefined;
     let state = {};
     let result = promote(state, null, null, issue);
```
oderby: cleanup: remove `issue` from call here and below.

static/js/components/promote.js
```diff
@@ -1,24 +1,33 @@
...
 
-module.exports = (state, prev, send, issue) => {
-  let url = encodeURIComponent("https://5calls.org?utm_campaign=twshare")
-  let additionalComps = encodeURIComponent("&via=make5calls")
+module.exports = (state, prev, send) => {
+  // share widgets for both platforms don't play well with dom manipulation so we'll make our own
+
+  // if no issue is selected, use these default parameters
+  let url = encodeURIComponent("https://5calls.org")
+  let additionalTwitterComps = "&via=make5calls"
```
oderby, line 9: Why remove the `encodeURIComponent` call for this arg?

nickoneill: `&` and `=` are escaped and it appears as text as opposed to part of the url like it should

static/js/components/promote_test.js
nickoneill: Yeah, these are failing too, they need a bit more love.

nickoneill:
OK, that might be the right way to fix those tests.

static/js/components/promote.js
oderby: ahh, yes, I didn't look close enough at how it was being used. Yes, you are correct.

oderby (approved):

Issue #227, opened 2017-02-27
Add some reasonable ESLint rules


Mr0grog:
Now that ESLint is part of the test process, we should actually add some reasonable code style rules to it. (At current, it’s only used to ensure people don’t leave `console` methods in their at PR time).

cdoremus:
+1 @Mr0grog. Could you suggest an ```.eslintrc``` file to start with (or set of rules that we can add to an ```eslintConfig``` field in ```package.json```)?

Mr0grog:
I think someone else who has actually contributed substantively to the JS codebase should choose a set of rules that best match the intended coding style here. I’m pretty sure I’m not that person; I haven’t examined much of the JS source in any detail and only made two very minimum-impact changes.

oderby:
As someone who's contributed a little bit more, but not much, I think it's safe to say there's not a lot of opinionated people around here that will object overly much to a sane recommendation. So, going off that...any reason not to just use the [recommended set of rules](http://eslint.org/docs/user-guide/configuring#using-eslintrecommended)? Turning those on and running on our code base only produces two types of errors: ["no-undef"](http://eslint.org/docs/rules/no-undef) and ["no-unused-vars"](http://eslint.org/docs/rules/no-unused-vars), both of which seem easy to fix/turn into warnings/turn off.

cdoremus:
@oderby based on your recommendations, I modified the .eslitrc.json file in the static folder to look like this:
```
{
  "parserOptions": {
    "ecmaVersion": 7,
    "sourceType": "module"
  },
  "env": {
    "browser": true,
    "commonjs": true,
    "mocha": true
  },
  "extends": "eslint:recommended",
  "rules": {
      "no-unused-vars": "warn",
      "no-undef":  "warn",
      "no-console": "error"
  }
}
```
I set the ```no-unused-vars``` and ```no-undef``` rules to issue a warning rather than being ignored (value === "off") because these things should be cleaned up as time allows.

Please comment.

nickoneill:
The rules specified here look totally reasonable. We can add/improve them as issues appear but this  is an excellent place to start.

cdoremus:
@nickoneill thanks for chiming in. I've assigned this to myself and will put together a PR in a day or two to give anyone else a chance to comment on my proposed changes.

cdoremus:
This issue has been implemented in PR #233

PR #233, opened 2017-03-06
Updated eslint rules per issue #227

cdoremus:
Changes to ```.eslintrc.json``` are exactly as I recommended in issue #227.

Please note that there are currently 117 warnings for ```no-undef``` and ```no-unused-vars``` violations that should be fixed as time allows.


static/.eslintrc.json
```diff
@@ -5,9 +5,14 @@
   },
   "env": {
     "browser": true,
-    "commonjs": true
+    "commonjs": true,
+    "node": true,
```
Mr0grog: We should only set `node: true` for scripts in the e2e directory (where `browser: false` should also be the case), since everything in this directory is _only_ intended to run in the browser.

cdoremus:
BTW, I checked out the impact of adding an ```indent``` rule, which determines how many spaces (or tabs) the code should indent. A two space indentation rule shows 98 violations in our code. 

The nice thing about the ```indent``` rule is that it is auto-fixable by using the ```-fix``` command-line option.

Still, this is something I think we should defer for now given the number of ```no-undef``` and ```no-unused-vars``` violations we have, which I think are more serious in that they decrease code clarity.

oderby:
@cdoremus thanks for taking this on. One quick note re:[removing console rule](https://github.com/5calls/5calls/pull/233/commits/202fe9523a23982fe0620d6e5daf61826fa6254f): we explicitly want that rule to be set, regardless of what defaults/presets we use in the future (see #201 for discussion). That is, although it is redundant for now, it might not be in the future. Further, it doesn't do any harm to have it redeclared in this way, and only helps reinforce that it is an intentional setting. Therefore, I'd like to see it remain (ideally with a comment...but not sure how...because json...)

oderby:
Also, agreed on not turning on `indent` rule yet. Another one we might consider turning on, after we decide what convention to follow, is the [`semi`](http://eslint.org/docs/rules/semi) rule, which determines if lines should terminate with semicolons or not. Our code is half and half (half semicolons ending statements, half omit the semicolon).

nickoneill:
I support this change! Is the idea to turn these on and bring the code to support these rules in this PR? Or just change the rules here?

cdoremus:
@nickoneill this change will setup rules based on eslint project recommendations (```eslint:recommended``` -- see http://eslint.org/docs/rules/). Fixing rules will need to be done in another PR (or PRs). 

Right now, the front-end code would fail due to violations of ```no-undef``` and ```no-unused-vars```, so I have set them to issue a warning rather than an error to signal that they should be fixed. Also, warnings  will not fail our CircleCI build (the ```test:ci``` Gulp task depends on the ```eslint``` task).

@oderby I'd be willing to put back the ```no-console``` rule, but I first want to hear from @Mr0grog about this since he is the official reviewer of this PR. Others are also welcome to chime in.

Mr0grog:
> I'd like to see it remain (ideally with a comment...but not sure how...because json...)

FYI, ESLint uses a looser JSON parser specifically so you can have comments in your .eslintrc files 😉 

I also agree that I think we’d like the `no-console` rule to be extra-explicit here—we are including it for more reasons that just good practice and really want to as clear as possible that we intend for `console` usage (in browser code) to fail the build. (To be clear, it’s totally fine to remove the rule for Node code, like the e2e tests. The main problem we’re concerned about here is compatibility, not spurious logging.)

👍 on adding `semi` as well. I am generally pro-semicolon, but am fine with whichever people want so long as we’re consistent.

static/.eslintrc.json
greggraf: If I am right, the eslint task in the gulpfile currently only lints the files in `static/` - should we consider adding `e2e-tests` as part of this work?

Mr0grog: Ah, that is a good point! We want to make sure it doesn’t run on the generated content in `/app`, but it really should be changed to run everywhere else.

cdoremus: @greggraf good catch! I just pushed that change.

gulpfile.js
```diff
@@ -174,8 +174,8 @@ gulp.task('test:ci', ['eslint'], function() {
 gulp.task('eslint', function() {
   const eslint = require('eslint');
   const linter = new eslint.CLIEngine();
-  const report = linter.executeOnFiles(['./static']);
-  
+  const report = linter.executeOnFiles(['./static', './e2e-tests']);
```
Mr0grog: Maybe better to execute on the project root and add an `.eslintignore` file (same sort of setup as `.gitignore` files) to ignore `app/*`? That way the linter will check things like this gulpfile and our karma configs, too.

cdoremus: Rob (@Mr0grog), I tried a bunch of different configurations and I've come to the conclusion that we should just have one root level ```.eslintrc.json```, which would contain ```"browser": true``` and ```"node": true``` "env" fields. I think that would make any future changes easier to do rather than trying to maintain multiple ```.eslintrc.json``` files. Do you have any objections to this?

BTW, running the ```estlint``` Gulp task using a root ```.eslintrc.json``` file shows this error in ```karma.ci.config.js``` : ```62:5  error  Duplicate key 'sauce_safari'  no-dupe-keys```
I'm not sure how to fix this error. Please advise.

Mr0grog: > we should just have one root level `.eslintrc.json`, which would contain `"browser": true` and `"node": true` "env" fields.

Did you know that `.eslintrc.*` files inherit in order to make the kind of maintenance you’re concerned about simpler? You can have a top-level one like:

```js
{
  "parserOptions": {
    "ecmaVersion": 7,
    "sourceType": "module"
  },
  "env": {
    "es6": true,
    // things at the root run in node, like karma conf, gulp, etc.
    "node": true
  },
  "extends": "eslint:recommended",
  "rules": {
      "no-unused-vars": "warn",
      "no-undef":  "warn",
      "no-console": "error"
  }
}
```

And then you only need to specify `"env"` in the configs for each directory…

`static/.eslintrc.json`:

```js
{
  "env": {
    "browser": true,
    "commonjs": true,
    // NOTE: it might be a good idea to move unit test files into a separate
    // directory so we don't have to do this
    "mocha": true,
    // turn node off in this directory because this content is all browser-only
    "node": false
  }
}
```

`e2e-tests/.eslintrc.json`:

```js
{
  "env": {
    "mocha": true
  }
}
```



> running the `estlint` Gulp task using a root `.eslintrc.json` file shows this error in `karma.ci.conf.js` : `62:5 error Duplicate key 'sauce_safari' no-dupe-keys`
> I'm not sure how to fix this error. Please advise.

Huh, that was a dumb move on my part: https://github.com/5calls/5calls/blob/master/karma.ci.conf.js#L62

You should just rename that second object to `sauce_safari_9` instead. As it is, the first object is never getting used!

Mr0grog: Side note: if it wasn’t clear, it doesn’t matter what those browser configuration objects in the Karma config file are named. They’re totally arbitrary as far as Karma is concerned; they only need to be unique.

cdoremus: Thanks, again, for educating me. Next time I will RTFM before I further tax your patience ;).

Mr0grog: No worries!

greggraf:
For what its worth, I think this is awesome:
 - linting the top level of the project
 - moving `eslintrc.json` to top level
 - taking advantage of eslint built in 'inheritence' mechanism

Mr0grog (approved):
Oh, geez, I hadn't realized I never clicked approve! Looks great.

cdoremus:
@Mr0grog sorry to be a bug up your butt, but the Review Required notice is still showing.

Mr0grog:
Yeah, apparently I no longer have rights for this? It won't let me approve.

Mr0grog (approved):
It’s ok, Github, let’s try this again.

Mr0grog:
I guess this needs @nickoneill

nickoneill:
For approval? Or merging?

nickoneill:
I tried adding you to the team as well, see if that fixes things.

Mr0grog:
Ha! Looks like that did it—just accepting the org invite caused this PR to suddenly be approved :P

Guess there’s no separation between merge rights and approval rights, which… well doesn’t seem totally crazy now that I think about it.

PR #237, opened 2017-03-09
JS Code Coverage

greggraf:
This pull requests is an attempt to introduce coverage to the client side codebase.

Changes have been made to the existing karma configuration file that gets run during the`test` and `test:watch tasks`, but not in the karma ci configuration.

Every time these tests are run a coverage report is generated in a top level `coverage/` directory.


nickoneill:
Thanks @greggraf! What does the change to the `rootNode` code in main.js accomplish here?

greggraf:
@nickoneill - currently `main.js` does not get included in the karma test process. 

When I have tried to include it (in this case to make it part of the coverage report, but also when I have tried to write unit tests for code in `main.js`) I get a JavaScript error at:
```
document.body.replaceChild(tree, rootNode);
```
 `rootNode.` is `null` because the DOM element with the id `root` does not exist in the Karma html file that runs the tests. 

The code here adds a check that `document.getElementById('root')` actually returns something before trying to use it.

nickoneill:
SGTM, definitely good to keep an eye on our coverage, possibly as an easy first step for new volunteers. Write a test!

nickoneill (approved):

Issue #239, opened 2017-03-10
In IE 10 / Windows 7,  the page throws an error and the application fails to start


greggraf:
One of the supported browsers (based on the Sauce configuration) throws the the following error when I load  https://5calls.org/
```
SCRIPT1002: Syntax error 
app.js, line 1 character 12164
```
This is happening because versions of IE before 11 do not support the ES6 `const` keyword and though the code is run through the es2040 transform to make it compatible with older browsers, the transform does not currently operate on external dependencies (code in the `node_modules` folder).

I would like to propose adjusting `gulpfile.js` and `karma.conf.js` to  apply this transform more broadly using `opts.global` described here: https://github.com/substack/node-browserify#user-content-btransformtr-opts as a fix.

greggraf:
The fix in #241 addresses the error thrown in IE 10, but uncovers more problems.

One is that the app starts but `startup` never gets invoked, so the app never makes its requests to the IP/location service or the 5calls.org backend. 

I suspect the reason is that `startup` gets invoked in a number of places by using an `onload` attribute of a <div> element. The <div> does not support the `onload` attribute in IE 10. In fact, technically HTML5 its only supposed to work on <body>, <frame>, <iframe>, <img>, <input type="image">, <link>, <script>, and <style> tags. 

Should IE 10 be a supported browser?
Should an alternative trigger be used for `startup`?
If yes, why do we trigger `startup` in so many places and not just when app.start() happens?

greggraf:
Looking further in to the `onload` issue, I have discovered new information

It appears that `onload` is provided by the bel module
https://github.com/shama/bel#lifecycle-events, which uses https://github.com/shama/on-load which relies on `MutationObserver` which is not supported natively in IE 10. 

There are at least three options:

- drop support for IE 10 (and remove from the configuration when tests are run on Sauce Labs)
- use this polyfill https://github.com/polymer/MutationObservers to allow  IE 10 to work
- avoid using `onload` and move `startup` out of unload in the various places we trigger in other ways

I lean towards the last, as my naive reading of the code leads me to believe `startup` can be called in fewer places, and does not need to be tied to rendering any particular part of the page. I lean towards triggering it on `app.start()` by adding it as a `subscription`.

nickoneill:
analytics says that about 100 IE 10 users have visited the site over the last 30 days. That's about 1% of the total Internet Explorer traffic to the site... which accounts for about 1% of the overall traffic to the site. That's a very small number.

[...]

PR #241, opened 2017-03-11
Globally apply es2040 for fuller browser support

greggraf:
This pull request addresses #239 - applying the ES2040 shim to lower dependencies (Choo was the culprit for the bug).


nickoneill:
Looking through this and #239 - it's not clear if this is still the solution to provide IE 10 support? Can we enable saucelabs testing on IE10 to be clear what this is fixing?

Please advise, I don't have a problem running es2040 on everything if it makes shit work.

greggraf:
I hope this clarifies and does not add to confusion (wincing smiley face)

1. IE 10 is already enabled for testing on SauceLabs
2. The error in question is:
```
SCRIPT1002: Syntax error 
app.js, line 1152 character 1
```
3. This error exhibits itself in IE10 on SauceLabs only on branches  where `main.js` is included in Karma testing (`main.js` is the only file that requires `choo`)
4. Visiting 5calls.org with IE10 also exhibits the error (visiting with Chrome on Windows does not throw an error)
5. I can replicate the error when running the app locally and using IE10 (Chrome on Windows does not throw an error)
6. For illustration: 

   This is a screenshot of the site using IE 10:
     https://www.webpagetest.org/result/170321_1C_MCW/1/screen_shot/
   
   This is a screenshot using Firefox:
      https://www.webpagetest.org/result/170321_Z3_ME8/1/screen_shot/

As best I can figure, the code in Choo.js uses `const` and ES2040 does not transpile dependencies in `node_modules` by default. It works in Chrome because `const` is supported, but in IE10 it is not.

This fix (branch) works in IE10 for me locally, but leads to further IE10 bugs (like `onload` not being supported) as I have started to list in #239

nickoneill:
Replied specifically about IE 10 in #239. We can still apply es2040 globally via this PR if you think it's worth it with or without IE 10 support.

oderby:
If I understand correctly, this will recursively apply the es2040 transform to all code pulled into our application via require statements? If so, I have a mostly theoretical/philosophical concern that this changes the upstream dependencies, such that the code we're using is not necessarily the code they've distributed, and could result in subtle bugs/issues/confusions in the future when comparing functionality observed on our site with expected behavior of the upstream dependencies.

Also, going one step further on what Nick said - is this the right thing to do, independent of IE 10 support? 

Nevertheless, I'm leaning in favor of this change, because I trust es2040 transform is idempotent (and so my first concern is purely hypothetical), and think the defensive wins (against future bugs like #239) is worth it, regardless of IE10 support or not.

greggraf:
I agree about altering upstream dependencies. And its messy. What if an upstream dependency relies on an ES6 feature that ES2040 doesn't support.

For what its worth,  the approach of applying the transform globally is what is recommended here: https://github.com/yoshuawuyts/choo/blob/f84ec43fa58508cc20fe537d752a14901339f0cd/README.md#choo--internet-explorer--safari

I was thinking - absent applying ES2040 globally, why bother with the ES2040 at all? If the purpose of the plugin is to polyfill a subset of unevenly supported ES6 features, but dependencies are exempt what is it really doing?

But then, I think in the end, its safer to apply the transform to dependencies as an extra layer of compatibility guarantee.

nickoneill:
First chance we'll get to try this out will probably be April 9, our JS error tracking resets and we'll be better able to measure the effect of this!

oderby:
@nickoneill so are you proposing we defer taking further action on this PR until after April 9?

nickoneill:
Yes, we'll be able to get a baseline error rate on the 9th, then we can make this change on the 10th or later and see how it changes the error rates.

nickoneill:
I'm pretty happy with our baseline errors on Sentry (working on getting more people invited to check that out). What else has to happen to move forward with this @greggraf ?

greggraf:
@nickoneill - Unless I am misunderstanding something, I think all that is needed to move forward is to merge this PR.

Does the error that IE throws show up currently in Sentry?
```
SCRIPT1002: Syntax error 
app.js, line 1152 character 1
```

nickoneill:
I don't see it in sentry, but I can't even filter by IE 10, I wonder if sentry even supports that browser. Let's merge!

nickoneill (approved):

PR #243, opened 2017-03-15
I18n

schraj:
Localization draft.

Notes:

1.  This uses the i18Next localization package.
2.   It uses the i18next-xhr-backend package to load the locale files from the server when requested.  It does this once it finds the users browserLanguage.  Therefore, I've wrapped the "startApp" function in the callback from the initialization process for the i18n cache.  Otherwise, it will not have the localized resources as it is starting up. 
3.   If there is html in the localized string(and this is also the default), it wraps the localized string in a span tag (configurable, can also request a div tag) and then uses choo to compile it into dom nodes(choo just passes it to yo-yo to do this) so that choo can consume it.  If you set an argument that you want "justText" it will skip this step and just do the localization.
4.  The footer that is on the Index.html page is not handled as a choo widget and so I had to handle it differently.  You'll see that I set the localized values on it manually after the i18n cache is initialized.  If we moved the footer to a choo widget, this would be cleaner.  However, it would require a little structural markup change.
5.  The privacy.html page is not done.  We can talk about it once we evaluate the rest of this.  It is also not a choo widget so would need different handling like the footer.
6.  The main code for the localization is in /js/utils/translation.js
7.  The tests and some usage hints are in translation_tests.js
8.  The locale files are in /locales/.  I made a very naive spanish translation file just for tests.  It will have to be redone by a real translator.
9.   I still have an issue on the noContacts.js widget where we need to pass a function to the localized string.  I'll continue working on that.
10. There is a new gulp task called "test:windows" that I'm just using to get karma to work on windows and not planning to merge as part of this.


schraj:
It appears that the process of rendering a string into a choo fragment is failing on a bunch of the sauce labs tests.  Looks like it is an IE issue.

greggraf:
@schraj I think the problem here is around how this code gets transpiled:
```
 ? "return html`<div>" + template + "</div>`"   
```
The ES2040 transform will take something like
```
 return html`<div>foo</div>`  
```
and transpile it into ES5 code because it recognizes the backtick pattern as something it can parse.

In your case the backtick pattern is inside a concatenated string plus variable expression. I am guessing that the transform doesn't recognize it as a pattern it can transpile.

Perhaps replacing this:
```
// wrap it in the choo literal.  It also needs to be in a single node.
 template = useDiv 
             ? "return html`<div>" + template + "</div>`"    
             : "return html`<span>" + template + "</span>`"    

 // Create a function from this string.  It will compile the string into code.
 const func = new Function("html", template);

 // evaluate/run the code.  Choo will render the text/html into dom nodes
 return func(html);
```
with something like this:
```
 template = useDiv 
   ? html`<div>${template}</div>` 
   : html`<span>${template}</span>`    

 return template
```

schraj:
Thanks @greggraf.  You're exactly right.  However, I have it like that because Choo won't compile any html that is in that "${template}" variable into html nodes.  This method of evaluating the string was to get around that.  However, yes, we are now running up against the fact that these strings are not transpiled.

I have another way of getting these strings built into dom nodes that a choo literal can consume as a variable that I'll have to go back to.  

Thanks for your feedback on the error. It was interesting to realize that this was the issue.

greggraf:
Ah! I get it. Without the string evaluation code, we get `&lt;` instead of `<`

At the risk of kicking up more dust  and assumptions, I tried this locally:
```
template = useDiv 
  ? html(["<div>" + template + "</div>"])
  : html(["<span>" + template + "</span>"])   

return template
```

It seems to work based on a quick and dirty unit test I threw together using the output of the existing approach. If this works for you, i could put a pull request together.

schraj:
Actually, instead of "<", we need an actual dom node to be created.  That is what the choo fragment will expect because it will then create a wrapping dom node out of itself.  

if template is initially = "<a>link</a>"

when we use a method without the template literal, we have to turn it into a dom node or v-dom node and pass that node as a variable rather than the html string.

If we use the template literal, choo will interpret the whole string, after our variable string is added in, and parse it all into a dom tree.

PR #245, opened 2017-03-15
Tests for startup() in main.js

greggraf:
This is a first pass at starting to write unit tests for `main.js`.

The test file `main_test.js` currently only tests the startup() function, validating the existing logic.

To make `main.js` testable, I 
- moved the model definition into its own variable
- exposed the model definition variable on the exports object so it can be required into the test
- and wrapped the `appendChild` logic in a test for whether `rootNode` exists (it doesn't when running in the karma environment)

output:
```
  startup function
    geolocation has been set, or there are no active issues
      ✔ should not invoke send
    allowBrowserGeo is true, locationFetchType is browserGeoLocation and geolocation is not set
      ✔ should invoke send with fetchLocationByBrowswer
    no active issues, and geolocation is set
      ✔ should invoke send with fetchingLocation, then fetchActiveIssues
    active issues, geolocation is not set, but address is set
      ✔ should invoke send with fetchingLocation, then fetchActiveIssues
    locationFetchType is ipAddress and geolocation is not set
      ✔ should invoke send with fetchLocationByIP

```


greggraf:
depends on #241 to pass IE 10 tests on Sauce

PR #249, opened 2017-03-16
Add null check and clean up

bernardwang:
## Summary
Leftover changes to [#208](https://github.com/5calls/5calls/issues/208)

Adds null check to `scrollIntoView.js` in case the element is invalid.
Removes last usage of old dependency.


schraj (approved):
Looks good.

cdoremus:
These small changes should be incorporated into your PR #250

PR #229, opened 2017-03-03
Adds initial call impact

bernardwang:
## Summary
For issue [#195](https://github.com/5calls/5calls/issues/195)

1. Saves completed calls (not skipped) in localstorage
2. Adds reset functionality
3. Adds initial impact component to ```done.js```, or the completed issue page.

Creates new state variable `userStats`. `userStats.all` is an array of all non-skipped calls, containing `issueid`, `contactid`, and `result`. The other fields for `userStats` are for quick and easy access of call `result`, assuming that information will be displayed in the future.
```
userStats = {
      all: [], 
      contacted: 0,
      vm: 0,
      unavailable: 0,
}
```

The total calls is being displayed on the completed issues page, and only shows the "impact" message if there has been successful calls. Currently, any clicks to "Made Contact", "Unavailable", or "Left Voicemail" are registered in stats, even if the issue is already marked complete.

## Testing
New tests in `stats-test.js` and updated tests in `done-test.js`.

Videos of functionality:
[Does not show impact if no calls were completed](https://i.gyazo.com/e7970cfb33f0d960c36f05434fa85055.mp4)
[Counts successful calls](https://i.gyazo.com/579fe12da99529870cdd4f86ffa0bc72.mp4)
[Stats persist after page refresh](https://i.gyazo.com/9457fe6728fa1265a0165089ffd15dd4.mp4)
[Reset functionality](https://i.gyazo.com/394c32eff279a6c5ef7aeff602b90507.mp4)

## Risks
Many of these decisions were made with assumptions, so much of this is subject to change. Specifically, knowing what stats should be displayed, and where in the UI/UX, would help with this alot! Just let me know what should be changed!

PR #250, opened 2017-03-16
Fix scrolling when selecting an issue from About or Done pages (issue #247)

### LINKED ISSUE #247: Scrolling does not occur when activating an issue from the About or Done page
Scrolling an issue into view uses the `#content` selector, which does not exist in `aboutView.js` or `doneView.js`. It looks like it is just trying to scroll to an element that doesn't exist in the view yet. 

I don't know for sure, but on initial inspection I don't think it was introduced by [#208](https://github.com/5calls/5calls/issues/208). 

[About Page](https://i.gyazo.com/e09ff294dbe3322f99020db920438167.mp4)
[Done Page](https://i.gyazo.com/abfe0557dc4a5d7a7c75eccdde77d278.mp4)

bernardwang:
##Summary
[#247](https://github.com/5calls/5calls/issues/247)
Not sure if this is the best way to do it, but it is very fast fix.

### DIFF
--- a/static/js/components/issues.js
+++ b/static/js/components/issues.js
@@ -1,5 +1,5 @@
 const html = require('choo/html');
-const scrollIntoView = require('scroll-into-view');
+const scrollIntoView = require('../utils/scrollIntoView.js');
 
 const issuesHeader = require('./issuesHeader.js');
 const issuesList = require('./issuesList.js');
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,8 +10,8 @@
   },
   "extends": "eslint:recommended",
   "rules": {
-    "no-unused-vars": "warn",
-    "no-undef":  "warn",
+    "no-unused-vars": "error",
+    "no-undef":  "error",
     "no-console": "error"
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ pip-delete-this-directory.txt
 
 # Unit test / coverage reports
 htmlcov/
+coverage
 .tox/
 .coverage
 .coverage.*
--- a/README.md
+++ b/README.md
@@ -1,26 +1,33 @@
 # 5calls
 
-This is where development for [5calls.org](http://5calls.org) happens. Please check the issue list and pull requests before starting work so we can ensure you're not duplicating work! We're all volunteers and want to treat the time you dedicate to the site with respect. Ping  [@make5calls](https://twitter.com/make5calls) on twitter with an email to get an invite to our Slack.
+This is where development of the [5calls.org](http://5calls.org) frontend happens. Please check the issue list, pull requests and contributor guidelines before starting work so we can ensure you're not duplicating work! We're all volunteers and want to treat the time you dedicate to the site with respect. Ping  [@make5calls](https://twitter.com/make5calls) on Twitter with your email to get an invite to our Slack.
 
-Currently this repo contains both frontend and backend development pieces.
+# Table of Contents
+* [Development](#Development)
+    * [Front End](#Front_End)
+    * [Application Server](#Application_Server)
+    * [Quality Assurance](#QA)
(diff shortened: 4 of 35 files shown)


schraj (approved):
This looks correct to me.

cdoremus (changes requested):
I tried running your 247-pr branch locally and you are missing the 'raf' dependency in package.json which causes the gulp build to fail.
Also, the 247-pr branch is based on your master branch, which you created (and changed) in PR #226. It -- and any other PR -- should be based on an up-to-date master branch refreshed from 5calls/5calls. To do this, add a remote branch to your fork pointing to the 5calls/5calls repo and use that branch to refresh your master branch before you create a branch for your PR. Since this is a small change, I'll let this issue go this time, but please use the proper procedure for any new PR. Thanks.

bernardwang:
@cdoremus I see, next time I will follow that procedure. I incorporated the small changes from , [#249](https://github.com/5calls/5calls/pull/249) but this PR is quite messy, would it be better to close it and redo it properly?

nickoneill:
Nah, we can squash those commits into a single one for clarity. Make sure to use that option when this is ready to merge @cdoremus

bernardwang:
Ok, let me know if theres anything else I should do for this PR. I believe this will need to be merged before [#229](https://github.com/5calls/5calls/pull/229).

static/js/components/issues.js
```diff
@@ -1,5 +1,5 @@
 const html = require('choo/html');
-const scrollIntoView = require('scroll-into-view');
```
oderby, line 2: Can we now remove the dependency on `scroll-into-view` from package.json now? (Assuming this is the last use of it, and we don't re-break everything again, i.e. #255 ?)

oderby:
Latest set of changes looks good to me - @cdoremus any objections to merging?

cdoremus:
@bernardwang @oderby I noticed that when you click on the About link in the footer that the page doesn't scroll to the top. It would be nice to fix that in this PR. Any objections?

oderby:
If @bernardwang thinks it's an easy fix and can get it up soon, then fine by me.

bernardwang:
I didn't find any more instances of `scroll-into-view` when I searched in the project, so it _should_ be safe to remove.

The footer links are just ordinary links, so I added an event listener to attach scrolling. It's easy, but perhaps a bit sloppy. Maybe in the future the footer could be a component rather than being hardcoded in `index.html`? Open to any other suggestions!

oderby:
Yeah, when I agreed with @cdoremus I didn't realize the footer wasn't part of a component. Also, in #251 (wip on i18n support) they are already doing some work to pull the footer into it's own component. Given both of these facts (which make fixing it now rather difficult/ugly, as you point out), I'm actually no longer in favor of increasing the scope of this PR to fix scrolling to top of about page. Sorry for flip-flopping here...

cdoremus:
@bernardwang I agree with @oderby that we should hold off on the About link scrolling for now.

bernardwang:
@oderby What about this option from [#229](https://github.com/5calls/5calls/pull/229#pullrequestreview-28828338)?

oderby:
Seems fine to me. Pulled it down and it seems to work fine @cdoremus thoughts?

Also, @bernardwang this doesn't rebase cleanly onto master branch, due to recent changes - can you merge master and resolve those conflicts and push them up? Don't worry about the mess of the history - as @nickoneill said above, we'll squash when merging this PR in, so it'll get all cleaned up.

bernardwang:
I pulled, but didn't seem to get any conflicts. Is this ok?
I should probably rebase more in the future haha, this history is pretty bad

cdoremus (approved):
Looks good! Thanks for you patience with our reviews. All commits will be squashed when the merge is done.

nickoneill:
Thanks @bernardwang! This fixes some issues needed for #256 as well, much appreciated!

PR #244, opened 2017-03-15
Fix undef/unused variable ESLint warnings across a number of files

greggraf:
This fixes all warnings.
(this work is towards issue: "Fix no-undef and no-unused-vars eslint violations" #238)

PR #251, opened 2017-03-16
I18n - Draft of Localization

schraj:
1. This uses the i18Next localization package.
2. It uses the i18next-xhr-backend package to load the locale files from the server when requested. It does this once it finds the users browserLanguage. Therefore, I've wrapped the "startApp" function in the callback from the initialization process for the i18n cache. Otherwise, it will not have the localized resources as it is starting up.
3. If there is html in the localized string(and this is also the default), it wraps the localized string in a span tag (configurable, can also request a div tag) and then builds it into dom nodes using "DOMParser" or a polyfill so that choo can consume it when passed to the choo template literal as variable. If you set an argument that you want "justText" it will skip this step and just do the localization.  Because I am now using this slightly more expensive method of transforming the string for choo consumption, it might make sense to identify strings that must be "JustText" and these could be passed with the "justText" parameter set to true so that they can be handled without a dom translation step.
4. The footer that is on the Index.html page is not handled as a choo widget and so I had to handle it differently. You'll see that I set the localized values on it manually after the i18n cache is initialized. If we moved the footer to a choo widget, this would be cleaner. However, it would require a little structural markup change.
5. The privacy.html page is not done. We can talk about it once we evaluate the rest of this. It is also not a choo widget so would need different handling like the footer.
6. The main code for the localization is in /js/utils/translation.js
7. The tests and some usage hints are in translation_tests.js
8. The locale files are in /locales/. I made a very naive spanish translation file just for tests. It will have to be redone by a real translator.
10. There is a new gulp task called "test:windows" that I'm just using to get karma to work on windows and not planning to merge as part of this localization work.


static/html/index.html
```diff
@@ -140,5 +139,17 @@ <h2 class="callcount">5 Calls will be right back.</h2>
       ga('send', 'pageview');
 
     </script>
+    <script>
```
beausmith: It'd probably be better to make components for any text strings in this file so that `${t.getText("foo")}` can be used. This DOM search/replace implementation is too fragile.

schraj:
Note that there will need to be some coordination with the backend when the localization of the issues is implemented.  This pull request does not address these things yet.  

The following items will need addressed:

1) What locales is the backend currently supporting? Just a simple api call from the backend returning an array of supported locales.
2) The frontend will need to pass the supported locale that it wants the issues returned in from the backend.
3)  The frontend will need to take the localized issue and pass it through the renderer(in translation.js) to make it consumable by choo(if we want to support html in issues, currently this is not happening in issues-html not being put into them).

static/html/index.html
schraj: Yeah, I agree with that.  It looks like all but the footer is replaced immediately upon app startup.  That "Main" part of the page seems to be there in case the app is broken.  If broken, the translation won't work either.

At the bottom it has this:
<h2 class="callcount">5 Calls will be right back.</h2>

I do have a version that pulls out the footer into a component.  That is a pretty straightforward change that requires some structural markup change.  I didn't add that to this because I wanted to minimize the impact of this one change.  However, I agree that it would be the better way to go and it is an easy thing to do.

schraj: @beausmith Since this is the first thing that jumped out at you, I think I'll go ahead and merge in my other version that has the footer as a component.  Might as well do it the right way all at once.

schraj:
@beausmith I added in the footer component(removed it from Index.html) so that it can be localized like the rest of the site.

static/js/components/outcomes.js
```diff
@@ -7,9 +8,9 @@ module.exports = (state, prev, send) => {
   const currentContact = issue.contacts[currentIndex];
 
   const contactsLeft = issue.contacts.length - (currentIndex + 1);
-  const callsPluralization = contactsLeft > 1 ? "people" : "person";
+  //const callsPluralization = contactsLeft > 1 ? "people" : "person";
```
beausmith: nit: delete dead code.

static/js/components/outcomes.js
```diff
@@ -7,9 +8,9 @@ module.exports = (state, prev, send) => {
...
 
-  const contactsLeftText = contactsLeft + " more " + callsPluralization +" to call for this issue.";
+  const contactsLeftText =  t.getText("outcomes.contactsLeft", { "contactsRemaining": contactsLeft}); //contactsLeft + " more " + callsPluralization +" to call for this issue.";
```
beausmith: nit: delete dead code comment.

static/js/utils/translation.js
```diff
@@ -0,0 +1,56 @@
...
+    // justText - if true, the object will be returned as just a text string and not turned into a choo fragment.
+    //      there is at least one use case where this is required(a placeholder in a text input cannot have text within a span)
+    // useDiv - if true, wrap the text in a <div> tag instead of a span.  This is not currently used, but may be required
+    //              with larger fragments such as we will get for issues.  By default(false), the localized text is wrapped in
+    //              a <span> tag.  
+    //              Note: if the third parameter is false, this fourth parameter will be ignored because the text will not be wrapped at all.
+    getText : (key, variableObject, justText, useDiv) => {
+        variableObject = variableObject || {};
+        justText = justText || false;
+        useDiv = useDiv || false;
```
beausmith:
```diff
+    getText : (key, variableObject = {}, justText = false, useDiv = false) => {
-    getText : (key, variableObject, justText, useDiv) => {
-        variableObject = variableObject || {};
-        justText = justText || false;
-        useDiv = useDiv || false;
```

Use [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)

beausmith (commented):
Fantastic work @schraj! This looks solid. 🇺🇸 :handshake: 🇲🇽 

Just a few nits/suggestions.

One wish-list (aka scope creep) thought: It would ease the translation process for translators if prose (multiple paragraphs of text) on the About, Hypothesis, etc pages be contained in a single translation key. Often this prose doesn't always map 1:1… there may be need for additional paragraph, or a different list of articles, etc. I was excited for the markdown option because this would be supported. But this is not critical for this PR. 

One other nit: I wish that the translation methods were simply `${t('key')}` without the `getText` part. If the logic to get the browser locale was place in it's own utils file… 😃

cdoremus:
@schraj please make sure this code does not have any linting warnings or errors by rebasing against the latest master and running `gulp eslint`. The latest PR merged into master (#244) fixed all undefined and unused variable lint violations.

beausmith:
@schraj - shower thought… the ability for the content team to markup the issue text using Markdown would be the simplest way to allow them to add basic formatting.

static/html/index.html
```diff
@@ -54,8 +54,8 @@
 
     <header class="logo__header" role="banner"></header>
 
-    <div id="root" class="layout">
-
+    <div id="root">
+      <div class="layout">
       <aside id="nav" role="contentinfo" class="layout__side">
```
oderby: please fix up indenting.

static/html/index.html
```diff
@@ -94,36 +94,37 @@ <h2 class="callcount">5 Calls will be right back.</h2>
     <footer>
       <div class="tinyletter__form">
         <form action="https://my.sendinblue.com/users/subscribe/js_id/2p22o/id/1" method="get" target="popupwindow" onsubmit="window.open('https://my.sendinblue.com/users/subscribe/js_id/2p22o/id/1', 'popupwindow', 'scrollbars=yes,width=800,height=600');return true">
-          <label for="email">Get email alerts once a week</label>
+          <label for="email" id="footer-emailLabel">Get email alerts once a week</label>
```
oderby: I'm not seeing how these ids are used? Why are these added ids needed?

static/locales/es.json
```diff
@@ -0,0 +1,118 @@
...
+      "personWithCount": "{{count}} persona",
+      "personWithCount_plural": "{{count}} gente"
+    },
+    "about": {
+      "title": "Alrededor de 5 Calls",
+      "whyCallingWorks": {
+        "title": "Por qué la llamada funciona",
+        "justificationForCalling": "",
+        "justificationArticlesListHeader": "No lo tome de nosotros:",
+        "article1":"<a href='https://www.nytimes.com/2016/11/22/us/politics/heres-why-you-should-call-not-email-your-legislators.html'>'Here’s Why You Should Call, Not Email, Your Legislators'</a> <span class='about__source'>NY Times</span>",
```
oderby: Why aren't the link texts translated?

static/js/components/about.js
```diff
@@ -1,37 +1,39 @@
...
-      <p>Our Android app is made by <a href="https://github.com/dektar">dektar</a>, gregliest and <a href="https://github.com/5calls/android/graphs/contributors">more supporters</a></p>
-      <p>Our issues, scripts and social team is <a href="https://twitter.com/charmbtrippin">@charmbtrippin</a>, <a href="https://www.instagram.com/sara.n.g/">sara.n.g</a>, amanda.l, <a href="https://twitter.com/djpiebob">@djpiebob</a>, <a href="https://twitter.com/theRati">@theRati</a>, <a href="https://twitter.com/Lhhargrave">@lhhargrave</a>, amanda.n, <a href="https://twitter.com/sagerke">@sagerke</a>, Eleanor Wertman and more contributors.</p>
-      <p>We want to make advocacy accessible. We hope 5 Calls will make it effortless for regular people to have a voice when it’s needed most.</p>
+      <h3 class="about__subtitle">${t.getText("about.whoIs5Calls.title")}</h3>
+      <p>${t.getText("about.whoIs5Calls.whyWeWorkOnIt")}</p>
+      <p>${t.getText("about.whoIs5Calls.broughtToYouBy")}
```
oderby: Looks like we lost a bunch of the contributors listed here?

static/js/components/call.js
```diff
@@ -15,13 +18,9 @@ module.exports = (state, prev, send) => {
...
       <div class="call_complete">
-        <h2 class="call__title">No calls to make</h2>
-        <p class="call__text">
-          This issue is no longer relevant, or the URL you used to get here was wrong. If you clicked a link on this site to get here, <a href="mailto:make5calls@gmail.com">please tell us</a> so we can fix it!
-        </p>
-        <p class="call__text">
-          Next choose a different issue from the list to make calls about.
-        </p>
+        <h2 class="call__title">${t.getText('call.title')}</h2>
+        <p class="call__text">${t.getText('call.reason', { contactEmail: constants.contact.email})}</p>
```
oderby: I find the key `call.reason` here to be a bit misleading - it's the key for the reason why there is no calls to make. Perhaps a better key would be `noCalls.reason`? (Then also would change `call.title` -> `noCalls.title` and `call.nextStep` -> `noCalls.nextStep`)

static/js/components/callcount.js
```diff
@@ -1,17 +1,24 @@
...
+  function callCountAsNumber(state){
     let calls = Number(state.totalCalls);
     // Handle undefined input.
     // Number(undefined) is NaN, while Number("") is 0
-    calls = isNaN(calls) ? 0 : calls;
+    return isNaN(calls) ? 0 : calls;    
+  }
+
+
+  function callCount(state) {
```
oderby, line 19: (perhaps for future work) Would it make sense to move this into the [`interpolation.format` option](http://i18next.com/translate/formatting/)? Right now, this is the only place we expect to have numbers >999, but eventually we might see them somewhere else. Something like the following is what I'm thinking.

```javascript
interpolation: {
  format: function(value, format, lng) {
    if (value instanceof Number) {
      // Number.toLocaleString() doesn't work on Safari 9 (see https://github.com/5calls/5calls/issues/197)
      if (window.Intl && typeof Intl.NumberFormat == 'function') {
        return value.toLocaleString();
      } else {
        // As a fallback, use a quick-and-dirty regex to insert commas.
        // When in doubt, get code from stackoverflow: http://stackoverflow.com/a/2901298/7542666
        return value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
      }
    }
    return value;
  }
}

File: static/js/components/hypothesis.js [hunk 11]
```diff
@@ -7,26 +9,23 @@ module.exports = (state, prev, send) => {
...
-        <li>uses your location to find your local representatives so your calls have more impact</li>
+        <li>${t.getText('hypothesis.feature1')}</li>
+        <li>${t.getText('hypothesis.feature2')}</li>
         </ul>
 
-        <h3 class="hypothesis__subtitle">Get the 5 Calls app:</h3>
+        <h3 class="hypothesis__subtitle">${t.getText('hypothesis.getApp')}</h3>
 
         <ul class="hypothesis__apps">
         <li><a href="https://itunes.apple.com/us/app/5-calls/id1202558609?mt=8"><img class="ios" src="/img/app-store.svg" alt="5 Calls on the App Store" /></a></li>
```
oderby: Should we care about translating alt-texts?

static/js/components/issues.js
```diff
@@ -1,12 +1,13 @@
...
+const t = require('../utils/translation');
 const scrollIntoView = require('scroll-into-view');
 
 const issuesHeader = require('./issuesHeader.js');
 const issuesList = require('./issuesList.js');
 
 module.exports = (state, prev, send) => {
 	function debugText(debug) {
-    return debug ? html`<a href="#" onclick=${resetCompletedIssues}>reset</a>` : html``;
+    return debug ? html`<a href="#" onclick=${resetCompletedIssues}>${t.getText("common.reset")}</a>` : html``;
```
oderby: do we really expect devs to be debugging who need es support?

static/locales/en.json
```diff
@@ -0,0 +1,121 @@
...
+    },
+    "hypothesis": {
+      "title": "Make your voice heard",
+      "p1": "Turn your passive participation into active resistance. Facebook likes and Twitter retweets can’t create the change you want to see. Calling your Government on the phone can.",
+      "p2": "Spend 5 minutes, make 5 calls.",
+      "p3": "Calling is the most effective way to influence your representative.   Read more about <a href='#about'>why calling works.",
+      "featuresTitle": "$t(common.5Calls):",
+      "feature1": "provides phone numbers and scripts so calling is quick and easy",
+      "feature2": "uses your location to find your local representatives so your calls have more impact",
+      "getApp": "Get the 5 Calls app:"
```
oderby: Should be `"Get the $t(common.5Calls) app:"`

static/js/utils/translation.js
```diff
@@ -0,0 +1,56 @@
...
+
+module.exports = {
+    // key - the localization key in the locale file
+    // variableObject - a json object having variables that will be interpolated into the localized string
+    // justText - if true, the object will be returned as just a text string and not turned into a choo fragment.
+    //      there is at least one use case where this is required(a placeholder in a text input cannot have text within a span)
+    // useDiv - if true, wrap the text in a <div> tag instead of a span.  This is not currently used, but may be required
+    //              with larger fragments such as we will get for issues.  By default(false), the localized text is wrapped in
+    //              a <span> tag.  
+    //              Note: if the third parameter is false, this fourth parameter will be ignored because the text will not be wrapped at all.
```
oderby: I think you mean "...third parameter is **true**..."

static/locales/en.json
```diff
@@ -0,0 +1,121 @@
...
+      "enterAnAddressOrZipCode": "Enter an address or zip code"
+    },
+    "call": {
+      "title": "No calls to make",
+      "reason": "This issue is no longer relevant, or the URL you used to get here was wrong. If you clicked a link on this site to get here, <a href='mailto:{{contactEmail}}'>please tell us</a> so we can fix it!",
+      "nextStep":"Next choose a different issue from the list to make calls about."
+    },
+    "noContact":{      
+      "oneOfTwoDistricts":"The location you provided could be in one of two congressional districts.",
+      "enterYourLocation":"<a id='lnkEnterAddress'>Enter your address</a> to identify your representative in the House.",
```
oderby: Can we pass the id to use for the anchor as an interpolated value? e.g. `... <a id='{{anchorId}}'> ...`. This would avoid unnecessary coupling between the code and the translation files.

oderby: Alternatively, can we pass in a reference to the function as an interpolation, and leave it as an onclick handler? Or does interpolation only work with strings and numbers?

static/js/utils/translation.js
oderby: Agreed, but this has a subtlety that if we switch to this, will need to change all the places in the current version of code where we call `t.getText(key, null, true)` to `t.getText(key, undefined, true)` in order to trigger the default value for `variableObject`.

static/js/utils/domParser.js
```diff
@@ -0,0 +1,46 @@
+/*
```
oderby: Can you add a comment on what this monkeypatch is doing and why we need it?

static/js/utils/translation_test.js
```diff
@@ -0,0 +1,171 @@
...
+    });
+
+    it('should return the localized phrase in a "span" tag when third argument to getText() is missing', () => {
+      let key = 'common.reset';
+      let expected = '<span>reset</span>';
+      let result = t.getText(key, null);
+      expect(result.outerHTML).to.equal(expected);
+    });
+
+    it('should return the phrase in a "span" tag when third argument to getText() is missing', () => {
```
oderby: I think you mean "...when third argument...is false"

static/js/utils/translation_test.js
```diff
@@ -0,0 +1,171 @@
+const t = require('./translation.js');
+const chai = require('chai');
+const expect = chai.expect;
+
+describe('translation', () => {
+  describe('arguments to getText method', () => {
+    it('should return the localized phrase in a span when no interpolation or pluraliation arguments are passed in the second, or options, argument', () => {
```
oderby: (test aesthetic nit) I think the tests should be structured as follows:

```javascript
describe('translation')
  describe('getText')
    describe('arguments')
    describe('simple localization')
    describe('pluralization and interpolation')
  describe('getLocaleFromBrowserLanguage')
```

static/js/utils/translation_test.js
```diff
@@ -0,0 +1,171 @@
...
+      let key = 'outcomes.contactsLeft';
+      let options = { 'contactsRemaining': 0 };
+      let expected = 'people';
+      let result = t.getText(key, options, true);
+      expect(result).to.contain(expected);
+    });
+  });
+
+  describe('when calling the getLocaleFromBrowserLanguage method', () => {
+    it('with an undefined argument, should get "en" locale', () => {
```
oderby: Please stick with standard format of "should [result] when [scenario]" for test cases.

static/test/setupUnitTests.js
```diff
@@ -5,3 +5,32 @@
...
+    }
+}
+
+const options = {
+    // turn on this flag to see if you're localization keys are not correct.  It will log a message
+    'debug': true,
+    // the localized data, adding directly to the cache
+    'resources': namespacedLocaleObject,
+    // the language we're using for all of the regular tests.
+    // There are some tests specific to localization where we reload the cache with another locale, at that point we re-initialize
```
oderby: Did you add any tests which use a different locale? I didn't notice any...

static/js/utils/translation.js
```diff
@@ -0,0 +1,56 @@
...
+        let doc = parser.parseFromString(template, "text/html");
+        let node = useDiv ? doc.querySelector("div") : doc.querySelector("span");
+        return node;
+    },
+
+    // Get the user's locale.  
+    // If the user's locale is not supported, we fall back to "en"
+    // Currently, the supported locales are hardcoded.
+    //  Shall we check the back end for supported locales? or keep a config file in the front end?
+    getLocaleFromBrowserLanguage : (browserLanguage) => {
```
oderby: Did you look at using the language detector i18next provides, [i18next-browser-languageDetector](https://github.com/i18next/i18next-browser-languageDetector)?

static/js/main.js
```diff
@@ -429,6 +433,26 @@ app.router({ default: '/' }, [
...
+// need to get the localization resource file before bootstrapping the app's rendering process
+i18n.use(XHR)
+    .init({
+    //'debug': true,
+    'lng': locale,
+    'backend': {
+      'loadPath': 'locales/{{lng}}.json'
+    },
+    'fallbackLng' : constants.localization.fallbackLocale
+}, () => {
```
oderby: can't this just be `init(options, startApp)` instead of `init(options, () => { startApp();})` ?

oderby (changes requested):
Left some comments. I'm definitely jumping in and missing some of the context from the previous iterations, so apologies if I'm asking something you've already addressed/resolved.

Also, I would second @beausmith request that it be just `${t('key')}` (as a small nit)

oderby:
Another thing that you'll need backend support for is routing to fetch the locale files (it *just works* on localhost because gulp is actually running a local static file server, but without backend changes it won't work on prod)

static/locales/en.json
```diff
@@ -0,0 +1,121 @@
...
+        "title": "Join us",
+        "contactInvite": "This project is <a href='https://github.com/5calls'>open source</a> and volunteer made. If you’d like to join us in developing useful tools for citizens, please get in touch via <a href='https://twitter.com/make5calls'>Twitter</a> or <a href='mailto:{{contactEmail}'>email</a>."
+      }
+    },    
+    "callCount": {
+      "callCountPhrase": "Together we’ve made {{formattedCallsTotal}} $t(common.call,{'count': {{totalCalls}}})"
+    },
+    "contact": {
+      "localOfficeNumbers": "Local office numbers:",
+      "busyLine": "Busy Line",
```
oderby: Should be "Busy line?" (missing question mark and not capitalized "line")

gulpfile.js
```diff
@@ -192,6 +206,7 @@ gulp.task('eslint', function() {
 });
 
 gulp.task('test', ['eslint', 'test:js-unit']);
+gulp.task('test:windows', ['eslint', 'test:js-unit-windows']);
 
-gulp.task('default', ['html', 'html:watch', 'html:serve', 'sass', 'sass:watch', 'copy-images', 'copy-images:watch', 'scripts', 'scripts:watch', 'extra']);
+gulp.task('default', ['html', 'html:watch', 'html:serve', 'sass', 'sass:watch', 'copy-images', 'copy-images:watch', 'scripts', 'scripts:watch', 'extra', 'locales']);
 gulp.task('deploy', ['html', 'sass', 'build-scripts', 'extra', 'copy-images']);
```
oderby: Also need to add `'locales'` task as a part of the deploy task.

static/html/index.html
schraj: good catch.  was from an earlier iteration.  I'll remove them.

static/js/components/about.js
schraj: May have been from a early iteration where I was trying things out.  I'll get the latest.

static/js/components/callcount.js
schraj: Yes, great point.  I had considered this but didn't want to increase the scope of this already large change.

static/js/components/issues.js
schraj: probably not, but was just a single item, so seemed a harmless piece of syntactic sugar.  However, if all dev work is discussed in english this is definitely not needed.  Will remove it.

static/locales/en.json
schraj: We can definititely do the first(pass the id).  I was not able to get it to recognize a function.  I'll look again, and do the first suggestion as a fallback option.

static/locales/en.json
schraj: done

static/js/utils/translation.js
schraj: done

static/js/components/issues.js
schraj: done

static/js/components/hypothesis.js
schraj: yes, i think we should.  added them.

static/js/components/call.js
schraj: makes sense. changed.

static/js/components/about.js
schraj: fixed.

static/locales/es.json
schraj: I was thinking these might be different articles per language because as it is, they will point to english language articles.  For now, I'll translate the links.

schraj: fixed.

schraj:
@beausmith @oderby  I've made some interim fixes but not ready for another review yet.  Will let you both know.

gulpfile.js
schraj: done

static/js/components/outcomes.js
schraj: done

static/js/components/outcomes.js
schraj: done

static/html/index.html
schraj: done

static/html/index.html
schraj: done

static/js/components/outcomes.js
schraj: removed the t.getText(...) and now it is just t.('key')

static/js/utils/translation.js
schraj: yes, I looked at it but we have the issue of having to check with the backend to see also what it will support before choosing what language to use.  That means we would have to write a "custom detector" that plugs into their detection framework.  We essentially end up doing it ourselves anyway. I have now added the use of local storage to cache the locale so we don't look it up each time.

static/js/utils/translation_test.js
schraj: done

static/js/utils/translation_test.js
schraj: fixed

static/js/utils/translation_test.js
schraj: fixed

static/js/main.js
schraj: yep, good point. fixed.

static/locales/en.json
schraj: fixed

beausmith:
Thanks for all the changes @schraj. I know it's quite a bit of work. Can't wait to see this get merged.

static/js/utils/domParser.js
schraj: I've removed the monkeypatch.  It was for supporting older browsers(IE10), but I misread the compatibility chart and IE10 is supported without the patch.  It looks like the rest of the browsers should be current enough with respect to this functionality.

static/test/setupUnitTests.js
schraj: added.

schraj:
@greggraf Have you seen this type of error on Circle CI before?  The first several of the browsers work correctly and the tests pass and then these three give the following errors.  You can see the full logs by going into the last 15 or so failed builds.  

`Safari 10.0.1 (Mac OS X 10.12.1) ERROR
  Disconnected, because no message in 30000 ms.
29 03 2017 04:41:59.801:ERROR [reporter.sauce]: ✖ Test Disconnected
29 03 2017 04:42:06.217:WARN [Edge 14.14393.0 (Windows 10 0.0.0)]: Disconnected (1 times), because no message in 30000 ms.
Edge 14.14393.0 (Windows 10 0.0.0) ERROR
  Disconnected, because no message in 30000 ms.
29 03 2017 04:42:06.220:ERROR [reporter.sauce]: ✖ Test Disconnected
29 03 2017 04:42:07.464:WARN [Safari 9.1.2 (Mac OS X 10.11.6)]: Disconnected (1 times), because no message in 30000 ms.
Safari 9.1.2 (Mac OS X 10.11.6) ERROR
  Disconnected, because no message in 30000 ms.
29 03 2017 04:42:07.465:ERROR [reporter.sauce]: ✖ Test Disconnected`

I've been pulling out my changes that went in since this started happening, but can't pinpoint what is causing it.  Also, is there a way to get more detailed logs from the failing VMs?

greggraf:
@schraj - I have not seen this error before. I ran the tests locally on Safari and had no problem. I think the next move to troubleshoot would be to see if SauceLabs has any clues, but that means getting access to the account. I think @nickoneill has access, maybe @cdoremus. It seems like the browsers are hanging, but its strange that its only these.

schraj:
@greggraf Thanks greg, First I'll try rolling back my commits to the point where it passed and start adding them back in one by one.  If that doesn't work, I'll start digging into the saucelabs info further.

schraj:
Even after rolling back changes to a point at which this pullrequest was passing the CI tests, the ci run is now failing while providing no information.  

I'm going to close this pull request and open a new one. I'll then add in the changes progressively to see if I can identify where the ci error begins happening.

schraj:
I'm reopening this pull request to test it with @oderby fix.

static/html/index.html
```diff
@@ -103,7 +106,7 @@ <h2 class="callcount">5 Calls will be right back.</h2>
       </div>
       <div class="colophon">
         <a href="https://github.com/5calls/5calls">
-          <i class="fa fa-github" aria-hidden="true"></i> Open Source
+          <i class="fa fa-github" aria-hidden="true"></i> openSource
```
oderby: Probably should stay "Open Source"

static/js/components/noContact.js
```diff
@@ -1,24 +1,34 @@
 const html = require('choo/html');
+const t = require('../utils/translation');
 
 module.exports = (state, prev, send) => {
+  function initializeFragment(targetId) {
```
oderby: Perhaps a better name would be `initializeLocalizedAnchors`?

static/js/components/noContact.js
```diff
@@ -1,24 +1,34 @@
...
-                    <p><a onclick=${(e) => enterLocation(e)}>Enter your address</a> to identify your representative in the House.</p>
+      const targetId = 'lnkEnterAddress';
+      return html`<div onload=${() => initializeFragment(targetId)}>
+                    <p>${t("noContact.oneOfTwoDistricts")}</p>
+                    <p>${t("noContact.enterYourLocation", {anchorId: targetId})}</p>
                   </div>`
     }
     else {
-      return html`<h2><a onclick=${(e) => enterLocation(e)}>Set your location</a> to find your representatives</h2>`
+      const targetId = 'lnkSetYourLocation';
```
oderby: Now that I look at this again, is there any reason we need to give them both a unique id? Instead, we could use a single unique class name for both of them, put a single onload callback on the top level div for this component (line 22/32 below), and use `document.getElementsByClassName` in `initializeFragments` above. Not a big deal either way, just removes some complexity here.

static/js/main.js
```diff
@@ -80,6 +78,19 @@ store.getAll('org.5calls.completed', (completed) => {
   completedIssues = completed == null ? [] : completed;
 });
 
+let cachedUserLocale = '';
+store.getAll('org.5calls.userlocale', (userLocale) => {
+  if (userLocale.length > 0) {
+    logger.debug("user locale get", userLocale[0]);
+    cachedUserLocale = userLocale[0];
+  } else {
+    const cachedUserLocale = userLocaleDetection(navigator.language || navigator.userLanguage);
```
oderby: remove `const` here - you already declared the variable in the `let` statement on line 81

static/js/main.js
```diff
@@ -424,12 +439,8 @@ app.model({
       }
 
       send('setUserStats', data, done);
-      
-      // This parameter will indicate to the backend api where this call report came from
-      // A value of test indicates that it did not come from the production environment
-      const viaParameter = window.location.host === '5calls.org' ? 'web' : 'test';
```
oderby: did you mean to remove this?

static/js/utils/localization.js
```diff
@@ -0,0 +1,19 @@
+const i18n = require('i18next');
+const xhr = require('i18next-xhr-backend');
+const constants = require('../constants');
+
+module.exports = {
+    start: (cachedUserLocale, cb) => {
```
oderby: nit: rename `cachedUserLocale` to just `userLocale` or such - doesn't matter here if it's cached or not.

static/js/utils/translation.js
```diff
@@ -1,56 +1,34 @@
...
-        let locale = find(constants.localization.supportedLocales, (l) => l === browserLanguage.substring(0,2))
-        if (!locale){
-            locale = constants.localization.fallbackLocale;
-        }    
-
-        return locale;
-    }
-}
\ No newline at end of file
+/*key - the localization key in the locale file
```
oderby: can you fix up the indentation here? And also use the more standard format for block comments:
```javascript
  /**
   * this is a block comment. 
   * wheee
   */
```

static/js/utils/userLocaleDetection.js
```diff
@@ -0,0 +1,22 @@
+// Get the user's locale.
```
oderby: Is there a reason this file can't be merged into utils/localization.js?

static/test/setupUnitTests.js
```diff
@@ -24,12 +24,14 @@ const namespacedLocaleObject = {
 
 const options = {
     // turn on this flag to see if you're localization keys are not correct.  It will log a message
-    'debug': true,
+    'debug': false,
```
oderby: Did you mean to check this in?

static/js/utils/localization.js
```diff
@@ -0,0 +1,19 @@
...
+        const options = {
+            //'debug': true,
```
oderby: nit: remove dead code

static/js/utils/translation_test.js
```diff
@@ -0,0 +1,182 @@
...
+
+    describe('change locales', () => {
+      it('should return the spanish localized string when the locale is changed to spanish', (done) => {
+        i18n.changeLanguage('es', () => {
+          let key = 'common.go';
+          let expected = 'Vaya';
+          let result = t(key, null, true);
+          expect(result).to.equal(expected);
+          // change the language back to english so the rest of the tests will be set for that language
+          i18n.changeLanguage('en', () => {
```
oderby: You can put this in an [`afterEach` hook](https://mochajs.org/#hooks) within just this suite (so insert it at line 91).

static/js/utils/translation_test.js
```diff
@@ -0,0 +1,182 @@
...
+            done();
+          });
+        });
+      });
+      it('should return the spanish localized string when the country is changed to spanish-mexican', (done) => {
+        i18n.changeLanguage('es', () => {
```
oderby: Shouldn't this be `'es-mx'`? Or something other than `'es'` - right now it's identical test to the one above.

oderby (changes requested):

static/js/components/noContact.js
schraj: good point. changed.

schraj:
@oderby  Thanks for the great reviews.  With such a big change, your reviews have been essential. I believe I've now addressed all of the items.  

The final items to consider:
1) Coordination with the backend
2) A close look over the running site by another set of eyes, preferably content person?  With so many changes, seem likely that small text changes might have slipped through.

gulpfile.js
```diff
@@ -161,10 +153,17 @@ function runKarmaTests ({singleRun, configFile} = {}) {
...
+var Server = require("karma").Server;
+var path = require("path");
+
 gulp.task('test:js-unit', function() {
   return runKarmaTests({singleRun: true});
 });
 
+gulp.task('test:js-unit-windows', function() {
+    new Server({configFile: path.join(__dirname, "/karma.conf.js")}).start();
+});
```
Mr0grog: Apologies for the drive-by comment here, but is this still needed now that #272 was merged?

static/js/components/noContact.js
```diff
@@ -2,10 +2,10 @@ const html = require('choo/html');
 const t = require('../utils/translation');
 
 module.exports = (state, prev, send) => {
-  function initializeFragment(targetId) {
-    let el = document.getElementById(`${targetId}`);
-    if (el) {
-      el.addEventListener("click", (e)=>enterLocation(e), false);
+  function initializeLocalizedAnchors(targetClassName) {
+    let els = document.getElementsByClassName(`${targetClassName}`);
```
oderby: Why the interpolation here? Couldn't it just be `document.getElementsByClassName(targetClassName);`?

static/js/components/impactResult.js
```diff
@@ -1,15 +1,15 @@
 const html = require('choo/html');
+const t = require('../utils/translation');
 
 module.exports = (state) => {
   const contactedCalls = state.userStats.contacted;
   const vmCalls = state.userStats.vm;
   const unavailableCalls = state.userStats.unavailable;
-
+  // Jeremy- TODO, when bernard has finalized his styling, need to handle clash with styling of span because I'm
```
oderby: impact code (#229) has merged (with a small copy update in #142) - Can you please pull in that copy change (just merge master again) and then can we fully resolve this TODO?

In particular, on the latest master, only part of the callSummaryText is bolded:
image

static/js/main.js
```diff
@@ -465,9 +477,15 @@ app.router({ default: '/' }, [
   ['/more', require('./pages/issuesView.js')],
 ]);
 
-const tree = app.start();
-const rootNode = document.getElementById('root');
+let startApp = () => {
```
oderby: So technically, this cb will receive an error, if one occurs, from loading the locale file from the server (due to server being broken or just network hiccup) - we may want to do something other than blindly loading the app in such a case. Otherwise, we get something that looks like:
image
For now, I propose that in such a case (`err !== undefined`), we do not start the app, falling back on the base index.html. And then file a new issue to investigate improving the behavior of the app if we fail to load the locale files.

static/js/constants.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  contact: {
+    email: 'make5calls@gmail.com',
+    github: 'https://github.com/5calls',
+    twitter: 'https://twitter.com/make5calls'
+  },
+  localization: {
+    supportedLocales: ['en', 'es'],
+    fallbackLocale: 'en',
+    loadPath: 'locales/{{lng}}.json'
```
oderby: This needs to be `loadPath: '/locales/{{lng}}.json'` in order to ensure we always load from the root of the server, not relative to current url (otherwise, if you refreshed when on url /issue/#issueid, it would try to load the locales from /issue/locales/en.json...

static/locales/en.json
```diff
@@ -0,0 +1,137 @@
...
+      "totalCallCountText": "You have made <span>$t(common.callWithCount,{'count': {{myTotalCalls}}})</span>!"
+    },
+    "issues":{
+      "whatsImportantToYou": "What’s important to you?",
+      "issuesListItemSummary": "$t(common.callWithCount,{'count': {{callsToMake}}}) to make"
+    },
+    "location": {
+      "gettingYourLocation": "Getting your location",
+      "invalidAddress": "That address is invalid, please try again",
+      "chooseALocation": "Choose a location",
```
oderby: Why did this copy change? On [current master](https://github.com/5calls/5calls/blob/master/static/js/components/issuesLocation.js), the text is "Enter and address or zip code" and has been that way for 2 months...

gulpfile.js
```diff
@@ -42,22 +43,9 @@ gulp.task('html:watch', function() {
...
-    // Allow the development server to respond to URLs defined in the front end application.
-    // Assume that any URL without a file extension can be handled by the client side code
-    // and serve index.html (instead of 404).
-  
-    if(!(path.extname(req.url))) {
-      req.url = "/";
-    }
-    next();
-  }  
-
```
oderby: Looks like you dropped commit [39af7ec7](https://github.com/5calls/5calls/commit/39af7ec757918ed01e0cb3403aa43a3353642a2f#diff-b9e12334e9eafd8341a6107dd98510c9) when resolving merge conflicts. Please add it back in.

static/html/index.html
```diff
@@ -1,61 +1,63 @@
 <!doctype html>
 <html lang="en-US">
-  <head>
```
oderby: Are you making any intentional, substantive changes to this file? If not, I'd prefer you make whitespace changes (if you really care so much) in a different commit, just to keep this commit a little clearer.

.eslintignore
```diff
@@ -1,2 +1,2 @@
 app/*
-coverage/*
+coverage/*
```
oderby: Can you drop this whitespace change from your commit?

oderby (commented):
Took a more detailed pass at the site running with your changes, overall looks good, but found a few small things to address (also, @Mr0grog comment needs addressing as well).

Also, I'm not sure what changes you think are needed on the backend. This PR works when I run it against the backend locally.

Beyond that, here's the followon/remaining work to be done before this should be merged:
- [ ] address review comments
- [ ] Do another merge from master (if needed)
- [ ] create issue to localize FAQ page
- [ ] create issue to localize Privacy page (?)
- [ ] Post in #i18n channel to get help/review for en.json strings (probably easiest just to upload both files and ask for a review)
- [ ] Probably a good idea to get someone else to take another look over entire page and copy, making sure we didn't miss any copy changes that landed since you started this PR.
- [ ] Someone should look over the site in spanish (perhaps after you get the translations reviewed) to make sure it all fits together logically.

schraj:
Thanks @oderby.  I'll make those fixes in the next day or so.

The one issue that this PR will have without coordination with the backend is if the user's browser  locale is "es"(spanish").  In that case, the site will be in Spanish(using my bad Spanish translation) and the Issues will be in English.

For a solution, and maybe this is what you are thinking, I could set the supported locales to be just "en"(English).  Then, it would always just fall back to that. That way, we could get the PR merged and then handle coordination with the backend and issues next.

static/js/constants.js
schraj: done

static/js/main.js
schraj: when I try the err!==undefined, that itself throws an error. 

However, when I throw an error from within localization.start or the callback, I get the fallback index.html page as desired.  How, exactly, are you getting the result you are showing?

static/locales/en.json
schraj: I'm just using the wrong key. fixed.

gulpfile.js
schraj: nice catch. done

static/js/main.js
oderby: huh? You've declared `err` as an argument, right? e.g.

```javascript
let startApp = (err) => {  
  // If we errored on initializing localization, then we won't have the
  // right content for most of the app. For now, just fallback to the default index.html in that case.
  if (err !== undefined) {
    return;
  }
  const tree = app.start();
  ...
```

schraj: ah, gotcha.  thought you were looking for something global.  you are thinking of specifically a localization failure.  good point.

oderby:
@schraj let's definitely see if we can't get you some help on reviewing the translations from i18n channel - it might take a while, so the sooner we get the request up there, the better.

Regarding spanish site with english issues: my understanding from @nickoneill was that was perfectly fine. It's what we're doing in the apps and will be the case until we can coordinate/support spanish issues.

static/js/components/impactResult.js
schraj: fixed

schraj:
@oderby I've done everything on the checklist except these two:

- Probably a good idea to get someone else to take another look over entire page and copy, making sure we didn't miss any copy changes that landed since you started this PR.

- Someone should look over the site in spanish (perhaps after you get the translations reviewed) to make sure it all fits together logically.


I put out a request for translation of the Spanish file on the slack channel.  Will let you know how that progresses.

nickoneill:
Excited to get this in! I'll take it for a spin in the next few days.

gulpfile.js
```diff
@@ -208,6 +213,7 @@ gulp.task('eslint', function() {
 });
 
 gulp.task('test', ['eslint', 'test:js-unit']);
+gulp.task('test:windows', ['eslint', 'test:js-unit-windows']);
```
oderby: Remove this

static/html/index.html
```diff
@@ -1,63 +1,61 @@
...
+    <meta name="theme-color" content="#0032FF">
+
+    <link rel="stylesheet" href="/css/style.css">
+    <script src="https://use.fontawesome.com/e87f1bc9b3.js"></script>
+  </head>
+  <body>
+
+    <header class="logo__header" role="banner"></header>
+
+    <div id="root" class="layout">
```
oderby: I believe you need to keep the change that split this div into two divs:
```html
  <div id="root">		
     <div class="layout">
```

Otherwise you get the footer twice.

oderby (changes requested):
two things lost/missed in the back and forth, then it seems good to me.

schraj:
@oderby I've added in the spanish translations provided by Eleanor from the i18N channel.  I'm still waiting on 2 new additions related to the new donate button, but I've added in some rough translations for them until we get the real ones.

oderby:
Cool. I think we're super close. @nickoneill have you had a chance to take this for a spin yet?

And finally, @nickoneill I'm inclined to say we push with the current spanish translations and then make fixes/updates to spanish copy once it's deployed and easier to get feedback on. (As opposed to requiring one of us to sit down with a Spanish speaker and run this PR/branch in front of them to get feedback before launching). Sound good?

oderby (approved):

nickoneill:
This is pretty awesome :)

I made a merge for some conflicts related to the location changes, fixed the footer for the non-js version of the page and pushed one new string: `location.changeLocation`. It's quite straightforward to change localized elements or add new ones, so this is everything we could want when it comes to supporting this. Nice work!

Definitely agree with @oderby that this should go live and then we can get improvement feedback.

nickoneill:
Actually I can't push anything that's not the conflict fix here since this is your own fork, this is the commit to add: https://github.com/5calls/5calls/commit/7096c09fe2be742e109f4fc73272aa8c091bfb49

Trying to figure out the test failure now. It's not loading the logs for me so I suspect it's just a failed test run that we get every once in a while. Rerunning now.

Edit: definitely funk on sauce's side: `Can not start Safari 9.0 (OS X 10.11)`, will try again shortly.

nickoneill:
OK! Tests are properly running again. Text change from the merge is making a location test fail.

nickoneill:
@schraj checking in here, do you want to manually pull that commit from my branch into this so tests are running again? Then we can unblock some other work that's waiting on this and make incremental improvements here.

schraj:
Sure, I'll do that today.

schraj:
@nickoneill It looks like your "schraj-i18n" had a change to the index.html page that removed everything from the footer(which only shows up if the app can't load) except the donate button.  Did you want for this to be merged too?  It is merged currently, but I'll remove it as if/when you let me know what you wanted to do with it.

nickoneill:
Yes - there are some other changes to the footer in the FAQ PR and now that it's part of the choo page itself, there's no sense in keeping the order/links up to date in two places. It only shows if the page fails to load and well... we have bigger problems if that's the case.

nickoneill:
Woo!

Issue #246, opened 2017-03-16
Create additional E2E tests


cdoremus:
I am proposing additional end-to-end (E2E) tests to aid our development and refactoring efforts. I will concentrate on the major ```5calls.org``` use cases.

Comments and suggestions are welcome.

The E2E tests will make use of the [Page Object Pattern](http://www.seleniumhq.org/docs/06_test_design_considerations.jsp#page-object-design-pattern) to help separate page-specific code from test code to buffer against test fragility when page content changes.

As a start, these use cases will be tested (with test cases listed below each):

1. Display home page 
a. After browsing to home page, assert that page has been rendered

2. Display Issue detail page
a. After an issue has been clicked in the sidebar, assert that the issue details page is rendered.

3. Display More Issues page 
a. After 'view more issues' linked has been clicked, assert that More Issues page is rendered.

4. Lookup Address
a. Click on the address link to render the address input text box (if it is not already rendered). Enter a zip code into the text box and click Go. Check that the same zip code is now rendered as the address link.

5. Display About page
a. After About link is clicked, assert that user navigates to About page

6. Display FAQ page
a. After FAQ link is clicked, assert that user navigates to FAQ page

These additional test cases will be deferred to another PR (or PRs) once a PR based on the above use cases have been merged into master:
2.b. After clicking a call result button (Unavailable, Made Contact, etc.)  on issue details page, assert that the number of calls decrements or the '1 more person to call for this issue' message no longer renders because all calls have been made.
3.b. Assert that any completed (green checked) More Issue maintains checked state after the More Issue page closes and is subsequently re-rendered.
3.c. Make sure that clicking on 'view more issues' maintains the state of which contacts were clicked for a particular issue

cdoremus:
Making progress and got the tests working using http://localhost:8000 as the base Url. Should that Url also work in CircleCI/Sauce Labs or should I configure the e2e tests to run using a different Url in CI?

nickoneill:
I'm not sure? Who would be best to ask here, @Mr0grog?

greggraf:
I found this:
  https://circleci.com/docs/1.0/browser-testing-with-sauce-labs/

Which indicates http://localhost:8000 could work.

I even took a pass at adapting that sample code to what I think would work for this project (temporarily replacing the existing test command):
   https://github.com/greggraf/5calls/blob/test-e2e-on-saucelabs/circle.yml

[...]

PR #252, opened 2017-03-20
Additional e2e tests

cdoremus:
These end-to-end tests are to implement issue #246. The can be run with ```npm run test:e2e```.

The tests are configured to run using http://localhost:8000/ by default, but if the environmental variable BASE_URL is set, the tests will run against that Url. 

CircleCI and Sauce Labs needs to configured to run these tests.


cdoremus:
@Mr0grog thanks for your input. If you have time, you certainly are invited to do the CircleCI/Sauce configuration. 

I'd like for this PR to be merged into master. That will allow devs to run the e2e tests locally as a smoke test on their changes before they push a PR. The CircleCI/Sauce configuration can then be done in a separate PR. @nickoneill (or anyone else) do you have any feelings about this?

Whoever does the review of this PR, please make sure that the tests run locally using `npm run test:e2e` after the server has been started (`npm start`) in a separate window.

cdoremus:
@Mr0grog thanks for your input in #246. If you have time, you certainly are invited to do the CircleCI/Sauce configuration.

I'd like for this PR to be merged into master. That will allow devs to run the e2e tests locally as a smoke test on their changes before they push a PR. The CircleCI/Sauce configuration can then be done in a separate PR. @nickoneill (or anyone else) do you have any feelings about this?

Whoever does the review of this PR, please make sure that the tests run locally using `npm run test:e2e` after the server has been started (`npm start`) in a separate window.

e2e-tests/about-page.js
```diff
@@ -0,0 +1,75 @@
+const webdriver = require('selenium-webdriver');
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
```
oderby: why do we need chai in the page-object?

e2e-tests/about-page.js
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Encapsulates characteristic text value from the About page.
+   *
+   * @returns About page text
+   * @see #getAboutPageTextElement
+   *
+   * @memberOf AboutPage
+   */
+  getAboutPageText() {
```
oderby: I'm not sure I see the point of this method residing here. It's not the actual text from the page - it's the expected text. And expected text belongs in the tests, not on the page-object.

e2e-tests/about.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * End-to-end tests for the About page.
+ */
+const test = require('selenium-webdriver/testing');
+const chai = require('chai');
+const expect = chai.expect;
+const chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
```
oderby: Can we move this configuration into a central place?

e2e-tests/about-page.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Encalsulates characteristic text element on the About page.
+ *
+ * @returns the About page element
+ *
+ * @memberOf AboutPage
+ */
+  getAboutPageTextElement() {
+    const aboutPageSelectorBy = By.css(this.aboutPageSelector);
+    this.driver.wait(until.elementLocated(aboutPageSelectorBy), 3000);
```
oderby: why are the timeouts different?

oderby: Also, should we also provide the optional message, for increased verbosity, as best practice? e.g. this would become `this.driver.wait(until.elementLocated(aboutPageSelectorBy), 3000, 'should find the about page title');`

e2e-tests/about-page.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+/**
+ * Obtains the Why Calling Works link element
+ *
+ * @returns the link element
+ *
+ * @memberOf AboutPage
+ */
+  getWhyCallingWorksLinkElement() {
```
oderby: Isn't this on the home page? Why is it on the about page object?

e2e-tests/about.js
```diff
@@ -0,0 +1,47 @@
...
+
+test.describe('about page', function() {
+  let page = undefined;
+  test.beforeEach(function() {
+    this.driver.get(url);
+    page = new AboutPage(this.driver);
+  });
+
+  test.afterEach(function() {
+    page = undefined;
```
oderby: what's the benefit of resetting page to undefined after each test?

e2e-tests/about.js
```diff
@@ -0,0 +1,47 @@
...
+    // verify that About page renders
+    const element = page.getAboutPageTextElement();
+    return expect(element.getText()).to.eventually.equal(expected);
+  });
+
+  test.it('should display about page when "About" link is clicked on home page', function() {
+    const expected = page.getAboutPageText();
+
+    // click on home page link
+    page.getAboutPageLinkElement().click();
```
oderby: I would expect that tests for behavior/functionality of links on the home page reside in tests for the home page, not the about page. But I can see how this could be subjective, and it raises the general question - where do e2e tests that span multiple pages go? e.g. if I wanted to add an e2e test for #216, where should that go?

oderby (commented):
Consider this a preliminary, I'm getting up to speed, round of high-level comments.

e2e-tests/about-page.js
cdoremus: @oderby you are right. I pushed up a commit that removes the chai references from the page object files. That's what I get for putting too much trust in `eslint` :smile:

e2e-tests/about-page.js
cdoremus: If the 'why calling works' link (or the About link) changes on the home page, a test in `about.js`  fails. Page objects encapsulate UI data for tests. In this case the `AboutPage` page object (in `about-page.js`) provides data for tests in `about.js`. This includes the links in the home page that navigate to the about page.

e2e-tests/about.js
cdoremus: @oderby you're right; designing tests can be subjective. Back button tests could be done from any page. In the case of #216, I would probably put it in `issues.js` since you are testing the behavior related to issues (or calls). 

If the test spans multiple pages, I would put the test in the file that is the focus of the behavior and use the page object in service of those tests. I don't think it is productive to have a test that just parades through the web site willy-nilly. They should try to replicate normal web site use cases.

e2e-tests/about.js
cdoremus: @oderby no good reason. I just pushed a commit that gets rid of that code.

e2e-tests/about-page.js
cdoremus: @oderby I did a lot of experimenting with timeouts at first because due to network flakiness and my inexperience with the `webdriverjs` API. Certain pages would be tempermental; the FAQ page being the top of the list. 

In hindsight, I'd say that we could set a default timeout; I think 3000 milliseconds should work. I'll create a constant in the config file and play around with it. Still, the FAQ page should be much higher. The About page can also be an outlier.

cdoremus: I completely missed the message argument in the docs (:-1:) . I'll work on adding messages to these `wait()` calls.

cdoremus: I've added a defaultTimeout constant to `e2e-tests.config.js` and set it to 3000ms in the latest commit and used it in all tests.

e2e-tests/about-page.js
Mr0grog: I kind of agree with @oderby here. I think the encapsulation we’re looking for is not about “UI data for tests,” but instead about “logic for interacting with [a particular type of] UI.” (At least this is what I’ve always understood the page object pattern’s intention to be.)

To that end, if you wanted to keep this information in the page object, I think what would make more sense is not `getAboutPageText()`, but something like `isAboutPage()` (which returns a boolean) or `assertIsAboutPage()` (which makes an assertion). Otherwise, I think the string best belongs in the test, rather than the page object.

e2e-tests/about-page.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+/**
+ * Obtains the About link element in the home page footer
+ *
+ * @returns
+ *
+ * @memberOf AboutPage
+ */
+  getAboutPageLinkElement() {
```
Mr0grog: I kind of feel like this (and, likewise, `getFaqLink()` and a few others) belongs somewhere else. It might make sense to have a `Navbar` or `Footer` object that encapsulates these. You could also put them on a base class of all pages.

Mr0grog: Separately, I think the `Element` suffix here is overly verbose (and inconsistent with `getFaqLink()` and `getViewMoreIssuesLink()`). Usually I’m all for extremely wordy and extra clear names, but `Element` seems redundant when you already have `Link` in the name.

e2e-tests/about-page.js
```diff
@@ -0,0 +1,73 @@
...
+    this.aboutPageText = 'ABOUT 5 CALLS';
+    this.aboutPageSelector = 'h2.about__title';
+  }
+
+/**
+ * Obtains the Why Calling Works link element
+ *
+ * @returns the link element
+ *
+ * @memberOf AboutPage
```
Mr0grog: I think most JSDoc implementations today are clever enough to determine `@memberof` information when you declare the method using class syntax, so this might be unnecessary noise. (I know it works pretty well in WebStorm, at least, since I’ve been using it lately.)

Relatedly, `@returns` should really have a type, e.g. `@returns {WebElementPromise}` (according to http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_ThenableWebDriver.html#findElement)

Mr0grog: Also related: you shouldn’t need to specify `@class` above the `class AboutPage {` declaration, either.

e2e-tests/faq.js
```diff
@@ -0,0 +1,35 @@
...
+
+      // Click on icon inside FAQ link
+      page.getFaqLink().click();
+
+      // verify that FAQ page is rendered
+      let element = page.getFaqPageElement();
+      return expect(element.getText()).to.eventually.equal(expected);
+    } catch( error ) {
+      /*eslint no-console: */
+      return Promise.resolve(error => console.log('Problem getting FAQ page', error))
```
Mr0grog: Did you test this? To the best of my knowledge, this is not how `Promise.resolve()` works. It returns an already-resolved promise that *resolves* (not rejects) to the value you pass in as an argument.

I think maybe what you want here is:

```js
error.message = 'Problem getting FAQ page';
return Promise.reject(error);
// or
return Promise.reject(new Error('Problem getting FAQ page'));
```

Or something along these lines. This all seems a little weird, though, since all the webdriver methods should return promises/thenables. I would have expected the promise-based expectation you have at the end to carry any rejections through. What are you expecting to throw synchronously here?

e2e-tests/support/webdriverjs-util.js
```diff
@@ -0,0 +1,83 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const LocalStorageKey = {
+  LOCATION: 'org.5calls.location',
+  GEOLOCATION: 'org.5calls.geolocation',
+  ALLOW_GEOLOCATION: 'org.5calls.allow_geolocation',
+  GEOLOCATION_TIME: 'org.5calls.geolocation_time',
+  GEOLOCATION_CITY: 'org.5calls.geolocation_city'
+}
```
Mr0grog, line 15: I wonder if there’s a reasonable place to create a module with this information that both the client and the tests can share, that way these strings are duplicated.

e2e-tests/support/webdriverjs-util.js
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Takes a screenshot and writes it to the local file system.
+ *
+ * @param {webdriver.WebDriver} driver an instance of Selenium web driver
+ * @param {string} fileName the name of the screehshot file (default==='screenshot.png')
+ * @param {string} screenshotPath the path to the screenshot file (default===process.cwd())
+ */
+function takeScreenshot(driver, fileName = 'screenshot.png', screenshotPath) {
```
Mr0grog: `saveScreenshot()` or `writeScreenshot()` or something might more clearly describe why/how this is different than `driver.takeScreenshot()`.

Mr0grog:
Hope I’m not piling on too much here. The high-level note I’d make is that I think the `*-page.js` files belong in the `support` directory, since they are tools to support your tests, not the tests themselves.

e2e-tests/support/webdriverjs-util.js
cdoremus: I'd like to have the localStorage keys in a object in a utility file, but I thought that would be beyond the scope of this PR, and I wanted to write the E2E tests without mucking with regular code. I was hoping that this utility would be part of the `main.js` refactor planned in #206.

e2e-tests/about-page.js
cdoremus: @Mr0grog Owen (@oderby) and I had a productive chat this weekend about E2E test design, and I'm working on refactoring the page object classes to better encapsulate content and behavior. I hope to push some commits in the next day or two.

e2e-tests/about-page.js
cdoremus: I will be moving this code into `home-page.js` with the refactor that I am working on.

e2e-tests/about-page.js
cdoremus: These JSDoc annotations were automatically added by vscode. I agree that `@class` and `@memberOf` are overkill and will remove them.

e2e-tests/faq.js
cdoremus: Thanks for catching this. I will make the changes you suggested.

e2e-tests/support/webdriverjs-util.js
cdoremus: I agree, this function needs a better name. I like your suggestion.

e2e-tests/about-page.js
Mr0grog: Oh, cool! :thumbsup:

e2e-tests/support/webdriverjs-util.js
Mr0grog: That seems pretty reasonable to me. Maybe worth throwing a `// TODO:` or `// FIXME:` comment in there to help ensure it actually gets done in the future, though.

gulpfile.js
```diff
@@ -164,6 +164,27 @@ gulp.task('test:e2e', function() {
...
+gulp.task('test-one:e2e', function() {
+  var testFile = process.env.TEST_FILE;
+  return gulp.src([
+    './e2e-tests/support/setupEndToEndTests.js',
+    `./e2e-tests/${testFile}`
+  ])
+  .pipe(mocha({
+    reporter: 'spec',
+    timeout: 6000
+  }));
```
Mr0grog: @cdoremus Did you know that Mocha has a built in `grep` option that lets you filter tests/test suites by name? It’s a string on the command line or a regex in the options hash, e.g:

```sh
$ mocha --grep 'location lookup'
```

```js
mocha({
  grep: new RegExp('location lookup')
})
```

You could even write the `test:e2e` task to parse it as an argument and pass it in:

```js
gulp.task('test:e2e', function() {
  const mochaOptions = {
    reporter: 'spec',
    timeout: 6000
  };

  if (process.argv.includes('--grep')) {
    const grepValue = process.argv[process.argv.indexOf('--grep') + 1];
    mochaOptions.grep = new RegExp(grepValue);
  }

  return gulp.src([
    './e2e-tests/support/setupEndToEndTests.js',
    './e2e-tests/{*,!(support)/*}.js'
  ])
    .pipe(mocha(mochaOptions));
});
```

So tests would be like:

```sh
$ gulp test:e2e --grep 'location lookup'
```

cdoremus:
@Mr0grog and @oderby I've pushed out refactors of the home, about and faq page tests and their associated page objects. Please take a look at them and comment on my approach at your earliest convenience. 

Stay tuned for further discussion.....

cdoremus:
@Mr0grog after a Slack discussion this weekend with @oderby (containing a lot of valuable feedback), I've decided to refactor what I had in `issues.js` and `issues-page.js` into:
1. `issuesList/issuesList-page.js` containing tests and page object related to the issues part of the sidebar (excludes location lookup) and the 'more issues' list.
2. `calls/calls-page.js` a page and page object focusing on the calls page displayed when an issue is clicked. I'm planning the next e2e testing iteration (PR) to focus on the call page buttons.

@Mr0grog and @oderby I'd appreciate some feedback on this approach.

I'm going to keep the location-related tests and page object pretty much the way they are (with cleanup) and renaming `location-lookup-page.js` to `location-page.js` for consistency, but that work will wait until after I push the `issueList` and `calls` code.

gulpfile.js
cdoremus: @Mr0grog thanks for the info. I looked at ways of using a command-line argument, what I found seemed too invasive for one Gulp test, so I settled on this approach. Since E2E tests can be pretty slow, I wanted to be able to run one test at a time during refactoring. Once all the tests are stabilized, I'll work on incorporating your approach into `gulpfile.js`.

oderby:
Still parsing your latest changes, @cdoremus. But in the meantime, I've been distracted by the discussion on selenium about how they're going to be [deprecating their current way of running/managing async tasks](https://github.com/SeleniumHQ/selenium/wiki/WebDriverJs#moving-to-asyncawait). Now, that's not scheduled to happen until 2018, and so I'm inclined to just ignore it, but just wanted to make sure we're all aware and agreed on deferring this concern until much later.

cdoremus:
@oderby I am aware of the changes coming to WebdriverJS. I think if we are careful about using page objects, and follow the [test pyramid](https://martinfowler.com/bliki/TestPyramid.html) and [not have a lot of e2e tests](https://testing.googleblog.com/2015/04/just-say-no-to-more-end-to-end-tests.html) we should be able to handle the upgrade.

e2e-tests/about-page.js
cdoremus: @oderby I added the message argument to all `webdriver.wait()` calls in the last commit push.

e2e-tests/support/webdriverjs-util.js
cdoremus: @Mr0grog made this change in latest commit push

e2e-tests/support/webdriverjs-util.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * Holds Selenium webdriverjs utilities.
```
oderby, line 2: I'm confused - as far as I can tell, nothing uses this...why is this here?

oderby:
@cdoremus I went a little deeper than I initially intended, and explored some alternative concepts for structuring our e2e tests, which I've pushed up and explained [here](https://github.com/cdoremus/5calls/commit/5d08704bd5815affa96611074950afc70ec20082). I think each change/point in the proposal is mostly independent of the others, so we can pick and choose which we like.

oderby:
Also, I found it interesting to peruse wordpress's [e2e test suite](https://github.com/Automattic/wp-e2e-tests/tree/master/lib/pages), which that same author contributes to.

e2e-tests/support/webdriverjs-util.js
cdoremus: I use the all the code in this file at least once to debug tests. Client references to this code would be deleted when I was finished debugging. Debugging E2E tests can be painful, so I would like to keep this file in our repo for future use.

oderby: Should I have been using this? Or are you using this when writing other e2e tests?

cdoremus: I use these utilities when I have an issue that I need to debug by saving a screenshot at a particular place in the test or to get/set a `localStorage` value.  After I am done with the debugging, I remove the utility calls. Your mileage may vary.

cdoremus:
@oderby I'm OK with the changes in your proposal except I don't understand number 2. If you assert that each page/component is loaded on page object (PO) construction, then each test function would need to instantiate the PO  after navigation has occurred rather than just instantiating it in `beforeEach()`. It seems like this introduces necessary code duplication. It also puts an assertion in the PO; I thought best practice was to keep assertions in the tests. I'd rather have an extra test that explicitly calls something like a `pageOrComponentExists()` function.

Maybe I'm missing something here? Please advise.

oderby:
Regarding your second question - verifying correct page exists in PO construction does seem to be part of PO best practices. Every resource I've found on POs, including seleniumHQ, encourage it:

> There is one, single, verification which can, and should, be within the page object and that is to verify that the page, and possibly critical elements on the page, were loaded correctly. - [SeleniumHQ Docs](http://www.seleniumhq.org/docs/06_test_design_considerations.jsp#page-object-design-pattern)

And it's recommended in [blog posts](https://www.pluralsight.com/guides/software-engineering-best-practices/getting-started-with-page-object-pattern-for-your-selenium-tests#are-you-ready-) by [others](https://watirmelon.blog/2016/03/15/ama-page-object-best-practices/), too.

That said, I think it's born from the Java roots of selenium, when it's running in a more synchronous environment and not on SPA. I think the biggest intended benefit is that, if you ensure the correct page is loaded when the PO is constructed, you don't need to be as defensive in your PO methods about elements existing before querying them. However, as we've both discovered, in javascript when testing a SPA, it seems like we always need to be defensive (which is why we both, independently, came upon the pattern of 'wait for element to exist, then get the element' in all PO methods). Given that we still seem to need to do that level of defensive coding within the methods, the gains of asserting in the constructor seem to be lessened. (As an aside, I think there's still something I'm missing about why this is so difficult/flaky in JS, and I suspect if we can figure this out, then we may still be able to be less defensive in the methods).

Perhaps it's important to consider the implications of proposal 2 in the context of using the rest of the proposed changes, especially number 3. That is, if the POs encapsulate navigating the app (via clicking on links on the pages), then they can just return the new PO expected to result from the navigation. If the PO doesn't assert on construction, then you'll have your tests littered with a bunch of `expect(page.isInitialized())` after every navigation call. Which, while explicit, is another form of necessary duplication. 

To summarize: Asserting on construction seems to be best practice. However, that best practice seems born from a more synchronous environment and I'm not sure we're seeing all the benefits it's supposed to provide. Given how our app behaves, I don't think adopting proposal number 2 alone makes sense. However, if we decide to adopt the other points, especially number 3, I think then number 2 simplifies things and reduces the amount of code we write in tests.

cdoremus:
@oderby I agree with what you are suggesting in number 3. But I'm not sure I agree that it's linked to number 2. Thinking again about my last comment, I don't think that it's important that we assert that we are on a page. It's more important that we assert that elements used internally in page object methods have been rendered. If the wait times out, then either the element is missing because the page it is on is missing or the element or its selector has changed (more likely). It may have be a best practice to assert that the page exists in the constructor; it just doesn't make sense to me, which is probably my fault.

BTW, I've used Selenium with Java off and on for a number of years for smoke testing and the wait API has always been a part of it. It is a consequence of network latency.

cdoremus:
I just pushed up changes suggested by @oderby (with linting fixes). I'm planning to do some additional changes including implementing stubbed tests and some refactorings, so please comment, but hold off on approving this PR just yet. Stay tuned....

e2e-tests/app-navigation.js
```diff
@@ -0,0 +1,108 @@
+/**
+ * End-to-end tests for home page
```
oderby: Comment needs to be updated

e2e-tests/support/base-page.js
```diff
@@ -0,0 +1,57 @@
+const webdriver = require('selenium-webdriver');
+const until = webdriver.until;
+const config = require('./e2e-tests.config');
+
+/**
+ * Page object for issues list related content,
+ * which includes the issues list in the
+ * sidebar and the list on the 'more issues' page.
```
oderby: Comment needs to be updated

e2e-tests/support/base-page.js
```diff
@@ -0,0 +1,57 @@
...
+
+    this.ensureInitialized();
+  }
+
+  ensureInitialized() {
+    this.driver.wait(this.isInitialized(), config.defaultTimeout,
+                     "Page-Object page never loaded");
+  }
+
+  isInitialized() {
```
oderby: Should add a comment explaining what we expect the isInitialized method to do. Something like
```javascript
/*
 * isInitialized should contain logic to determine that the page/container under test is fully loaded and initialized.
 * Typically, this is done by looking for one or more elements to be present on the page and contain some expected
 * text. Every page object extending this base class needs to implement this method to ensure proper functioning
 * of the page object.
 * 
 * @returns {Promise<boolean>} A Promise which will resolve to true iff once page is loaded and initialized.
 */
```

e2e-tests/support/faq-page.js
```diff
@@ -22,9 +27,8 @@ class FaqPage {
    * from the FAQ page.
    */
   getFaqPageElement() {
-    const selector = By.css(this.faqPageSelector);
-    this.driver.wait(until.elementLocated(selector),
-      config.defaultTimeout * 2, 'FAQ page element not found');
+    const selector = By.css('header.page-header > h1');
```
oderby: With this change, this method should be `getFaqPageHeaderElement` now..

e2e-tests/support/faq-page.js
```diff
@@ -22,9 +27,8 @@ class FaqPage {
...
+    this.driver.wait(until.elementLocated(selector), config.defaultTimeout * 2);
```
oderby: Could clean this up a bit and use `return this.waitForElement(selector, 'FAQ header element not found');` here

e2e-tests/support/faq-page.js
```diff
@@ -22,9 +27,8 @@ class FaqPage {
    * from the FAQ page.
    */
   getFaqPageElement() {
```
oderby: Is this method even still used? If so, my 2 comments below apply. If not, we should remove it.

e2e-tests/support/home-page.js
```diff
@@ -37,35 +29,32 @@ class HomePage {
...
    * about page.
    */
-  getAboutPageLink() {
-    const aboutPageLinkTextBy = By.linkText(this.aboutPageLinkText);
-    this.driver.wait(until.elementLocated(aboutPageLinkTextBy),
-    config.defaultTimeout, 'About page link not found');
-    return this.driver.findElement(aboutPageLinkTextBy);
+  followAboutPageLink() {
+    const aboutPageLinkSelector = By.partialLinkText("About");
+    this.waitForElement(aboutPageLinkSelector).click();
```
oderby: I forgot to copy over your wait messages in all these locations - think you can do that?

e2e-tests/support/faq-page.js
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Tests that faq page element with
+   * known text is present.
+   *
+   * @returns {Promise<boolean>} resolves to true if the
+   * faq page element with known text can be
+   * found; otherwise false.
+   *
+   */
+  isFaqPage() {
```
oderby: Is this still being used?

e2e-tests/support/location-page.js
```diff
@@ -0,0 +1,83 @@
+const webdriver = require('selenium-webdriver');
+const By = webdriver.By;
+const until = webdriver.until;
+const config = require('./e2e-tests.config');
+
+/**
+ * Page object for location-related content.
+ *
+ */
+class LocationPage {
```
oderby: Needs to be updated to use new style - extend BasePage and implement `isInitialized`. Might also change over to use `this.waitForElement` helper if you want.

oderby (changes requested):

Mr0grog:
Hey, I just wanted to pop in here quickly and say what I didn’t before: I’m happy to help with the CI configuration for this, but I think it should probably be a separate PR that comes after this; there’s a lot of churn here and I think getting the final result loaded and running in CI should wait until this is settled.

e2e-tests/app-navigation.js
cdoremus: Updated in latest push

e2e-tests/support/base-page.js
cdoremus: Updated in latest push

e2e-tests/support/base-page.js
cdoremus: Updated in latest push

e2e-tests/support/faq-page.js
cdoremus: No longer part of the `FaqPageObject` class in my latest push.

e2e-tests/support/faq-page.js
cdoremus: No longer part of the `FaqPageObject` class in my latest push.

e2e-tests/support/home-page.js
cdoremus: Added to latest push

e2e-tests/support/location-page.js
cdoremus: Working on this. Stay tuned...

cdoremus:
@oderby and @Mr0grog I just pushed changes to `app-navigation.js` that fills in most of the stubbed e2e tests. Please review to make sure I am on the right track with these. In the meantime, I'll work on refactoring `location.js` and cleaning up the other code.

gulpfile.js
Mr0grog: So… now that you added `--grep` argument support to `test:e2e`, we should get rid of this task, right?

cdoremus: Yes, I just pushed a commit that does just that (gets rid of `test-one:e2e` task).

e2e-tests/app-navigation.js
```diff
@@ -56,20 +58,53 @@ test.describe('Should be able to navigate', function() {
...
+    test.it("to home page", function() {
+      page = new IssuesListPage(this.driver);
+      page.followFirstIssue();
+      this.driver.navigate().refresh();
+      page.followHomeLink();
+    });
+    test.it("to another issue page", function() {
+      page = new IssuesListPage(this.driver);
+      // navigate to 1st issue
+      page.followFirstIssue();
```
oderby:
```javascript
      page = new IssuesListPage(this.driver);
      // navigate to 1st issue
      page.followFirstIssue();
```
Maybe we should move these two lines into a `beforeEach` for this test suite ("from issue page")?

e2e-tests/app-navigation.js
```diff
@@ -56,20 +58,53 @@ test.describe('Should be able to navigate', function() {
...
+      this.driver.navigate().refresh();
+      let callsPage = new CallsPage(this.driver);
+      callsPage.getCallTitle().getText().then(text => {
```
oderby: should return this promise or take a `done` cb and call it after the nested expect, I believe.

e2e-tests/app-navigation.js
cdoremus: Fixed in latest push. More updates to come...

e2e-tests/app-navigation.js
cdoremus: Fixed in latest push. More updates to come...

cdoremus:
@oderby and @Mr0grog these are all the changes and updates I want to make in this PR (except for additional requests from either of you), so please review at your earliest convenience. All nits are welcome (I think?!?).

e2e-tests/app-navigation.js
oderby: Still unresolved. Should return this promise here.

e2e-tests/app-navigation.js
```diff
@@ -0,0 +1,143 @@
...
+    test.it("to another issue page", function() {
+      // navigate to 1st issue
+      page.followFirstIssue();
+      this.driver.navigate().refresh();
+      let callsPage = new CallsPage(this.driver);
+      callsPage.getCallTitle().getText().then(text => {
+        // navigate to 2nd issue
+        page.followIssue(2);
+        const callsPage2 = new CallsPage(this.driver);
+        return expect(text).to.not.equal(callsPage2.getCallTitle().getText());
```
oderby: I believe this is comparing a string to a promise (which is what getText() returns), so this isn't doing what we want. Instead, probably want something like:

```javascript
return expect(callsPage2.getCallTitle().getText()).to.eventually.not.equal(text);
```

e2e-tests/support/home-page.js
```diff
@@ -0,0 +1,96 @@
...
+    return this.isHome();
+  }
+
+  /**
+   * Checks that the current document is the home page.
+   *
+   * @returns {Promise<boolean>} resolves to true if the
+   * Home page exists containing the proper title text.
+   */
+  isHome() {
```
oderby: Why do we need this extra method? The only place which uses it is the above `isInitialized` method.

e2e-tests/support/location-page.js
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Displays the location text box after clicking
+   * on the location button.
+   *
+   */
+  displayLocationInputBox() {
+    const selector = By.css('button.subtle-button');
+    this.driver.wait(until.elementLocated(selector),
+      config.defaultTimeout, 'Location button not found');
+    const addressButton = this.driver.findElement(selector);
```
oderby: Can simplify this to 

```javascript
const addressButton = this.waitForElement(selector, 'Location button not found');
```

oderby (changes requested):
A few small things to fix, but after a cursory once-over, I think this is looking good.

As we get closer to merging this, we'll have to pay attention to what other changes are likely to merge soon: 
* #251 could necessitate some small changes, but probably nothing at all. 
* #287 will necessitate some updates to the location PO, but that's expected since it's redoing the page UI. Given that that PR is just starting out, I'd prefer to merge this one first and fix up the PO afterwards (in #287 or in a separate PR). That would also give us the benefit of using e2e tests to ensure continued functionality.
* I think we should hold off on #285 until this lands, and update any tests needed in that pr.
* #286 is adding new content, so should be largely independent. It's a separate question if we even should add e2e testing for that PR - it's pulling townhall data from a third-party db, so it'll have problems with mocking/reliability, but probably should have some e2e smoke test to ensure it's functioning. Anyways, I think we can safely defer conversations on e2e tests to that PR.

In summary, I think we may want to ensure that this is compatible with #251, as I expect that will (hopefully) merge soon.

e2e-tests/app-navigation.js
cdoremus: Fixed in latest push

e2e-tests/support/home-page.js
cdoremus: This was a leftover from an earlier aborted refactor. Moved code back into `isInitialized()` in latest push.

e2e-tests/support/location-page.js
cdoremus: Fixed in latest push

oderby (approved):

Issue #240, opened 2017-03-10
Consider turning on semi and indent ESLint rules


oderby:
As discussed in #233, we might want to turn on [`indent`](http://eslint.org/docs/rules/indent) and [`semi`](http://eslint.org/docs/rules/semi) rules for ESLint, to help enforce code quality in our codebase. This issue is to track this suggestion, and open up for discussion what standards we should enforce with respect to indentation and semicolon usage. Once we agree on a direction, we can go ahead and implement it, either alongside #238 or afterwards.

I propose we require an indent of 2 spaces (right now we're inconsistent between 2 and 4) and semicolons everywhere, except in one line blocks. This would translate to eslint settings as follows.
```javascript
{
  "indent": ["error", 2],
  "semi": ["error", "always", { "omitLastInOneLineBlock": true }]
}
```

Note that the `indent` rule documentation presents a long tail of corner cases to consider, but the default settings seem reasonable to me and don't seem to present much of an issue in most of our code base. That said, I'm largely unopinionated about how we hanlde the corner cases, so if anyone has strong feelings, feel free to suggest settings for them. 

For reference, the list of errors eslint flags with the above settings in our current code base is below. Running eslint with the `--fix` flag fixes all of them.

```text
~/code/5calls/web/e2e-tests/support/e2e-tests.config.js
  3:2  error  Missing semicolon  semi

~/code/5calls/web/static/js/components/about.js
  38:2  error  Missing semicolon  semi

~/code/5calls/web/static/js/components/call.js
  31:56  error  Missing semicolon  semi
  33:42  error  Missing semicolon  semi
  54:2   error  Missing semicolon  semi

~/code/5calls/web/static/js/components/callcount.js
  24:2  error  Missing semicolon  semi

~/code/5calls/web/static/js/components/contact.js
   8:57  error  Missing semicolon  semi
  11:13  error  Missing semicolon  semi
  16:19  error  Missing semicolon  semi
  20:6   error  Missing semicolon  semi
  31:8   error  Missing semicolon  semi
  37:49  error  Missing semicolon  semi
  39:16  error  Missing semicolon  semi
  56:2   error  Missing semicolon  semi

~/code/5calls/web/static/js/components/content.js
  23:2  error  Missing semicolon  semi

~/code/5calls/web/static/js/components/done.js
   8:5  error  Expected indentation of 2 spaces but found 4  indent
  30:2  error  Missing semicolon                             semi

~/code/5calls/web/static/js/components/hypothesis.js
[...]

PR #253, opened 2017-03-21
ran `eslint --fix` on all js files to fix tabs and semicolons

greggraf:
This pull request address "Consider turning on semi and indent ESLint rules" #240.

I updated the config in `.eslintrc.json` to turn on the rules for semicolons and indents, then ran the command `./node_modules/.bin/eslint --fix ./` to automatically fix any errors found.

It touched/fixed a large number of files and though it is a relatively simple change, it may impact the merge-worthiness of existing pull requests and branches.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,8 @@
   "rules": {
     "no-unused-vars": "error",
     "no-undef":  "error",
-    "no-console": "error"
+    "no-console": "error",
+    "indent": ["error", 2],
+    "semi": ["error", "always", { "omitLastInOneLineBlock": true }]
   }
 }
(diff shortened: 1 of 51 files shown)


oderby (commented):
Let's wait for #229 #256 and #250 to merge first, then we can re-do this change and push it in then. But otherwise, the changes look fine to me.

cdoremus:
I agree with @oderby that we should wait on this. I'm not sure how far along is #251 but we might want to wait for that one too.

nickoneill:
I'd be happy to get these improved rules turned on now if you think the time is appropriate @cdoremus / @greggraf

cdoremus:
I think I'd wait to use `eslint --fix` on our code until after #251 gets pushed to master since that PR involves a lot of changes.

oderby:
just pushed up an update, so this is ready to go if we want.

nickoneill:
👍 Will merge this tonight!

nickoneill (approved):

nickoneill:
Consistency 😍

PR #256, opened 2017-03-23
Changed urls to remove hashes.

nickoneill:
This change modifies the links to point to non-hash urls. I think I got all of them. We're also routing the more issues page to `/more` rather than `/issues` so we don't conflict with the api path.

When running the client with `gulp`, this will route to appropriate urls but reloading those urls will result in 404s. I'm not sure if we can modify the gulp script for testing to do a better job here?

In production, the go backend will automatically route this links correctly, even when reloading. This change is live now and you can remove any hash from links to see the correct behavior.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -374,7 +374,7 @@ app.model({
       } else {
         scrollIntoView(document.querySelector('#content'));
         store.add("org.5calls.completed", issue.id, () => {})
-        send('location:set', "/#done/" + issue.id, done)
+        send('location:set', "/done/" + issue.id, done)
         send('setContactIndices', { newIndex: 0, issueid: issue.id }, done);
       }
     },
@@ -410,7 +410,7 @@ app.model({
       // Use Choo's internal model to control Window.location. Fixes issue #161
       // For more information, see: https://github.com/yoshuawuyts/choo/blob/f84ec43fa58508cc20fe537d752a14901339f0cd/README.md#router
       // this strips the query string which breaks hashes, so temp workaround
-      send('location:set', "/#issue/" + data.id, done)
+      send('location:set', "/issue/" + data.id, done)
       // location = location.origin + "#issue/" + data.id;
       // location.hash = "issue/" + data.id;
     }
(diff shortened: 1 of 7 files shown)


cdoremus (changes requested):
You missed the #about href in the 'Learn why calling' link on line 18 in done.js

oderby:
looks fine to me, after fixing the issue @cdoremus raised.

oderby:
Also, how does this relate to #225 ?

nickoneill:
Uh, I guess I forgot I did #225? It looks like this has a few more spots than that does, so I'll start here and add the link @cdoremus pointed out and move from `setlocation` to standard href links.

static/js/main.js
```diff
@@ -410,7 +410,7 @@ app.model({
       // Use Choo's internal model to control Window.location. Fixes issue #161
```
greggraf: Since hashes are no longer used in the urls, I am wondering if it is still necessary to set the window.location here.

greggraf:
@nickoneill - maybe its too much of an edge case, and maybe this hack is too naive, but this is a possible solution for getting 404s on reload on non root url requests

Gulp development server respond with index.html to all requests #258

oderby:
@greggraf you're right, this won't be needed. I have an outstanding PR that
will need to be rebased after this lands -
https://github.com/5calls/5calls/pull/221/files#diff-93471d4bf84d0569fadcebe5fba72773L377
However, I would rather hold off on further increasing the scope of this PR
and leave fixing up history/navigation to a separate PR (the one I have
will be easy to fix up and use for that purpose).

nickoneill:
Fixed @cdoremus! And some small href adjustments.

nickoneill:
This PR makes the core issue in #250 more prominent (breaking the page) so that should be merged before this one is pulled in.

nickoneill:
Rebased now that #250 is merged. Let me know if there are additional changes to make.

oderby:
Looks good to me!

PR #258, opened 2017-03-24
Gulp development server respond with index.html to all requests

greggraf:
added middleware to gulp task to serve index.html to all localhost:8000 requests without an extension

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,7 @@ var gulp = require('gulp')
   , connect_logger = require('connect-logger')
   , spawn = require('child_process').spawn
   , mocha = require('gulp-mocha')
+  , path = require('path')
   ;
 
 var SRC = {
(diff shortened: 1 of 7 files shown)


oderby:
@greggraf can you clarify the purpose of this PR? This change won't affect prod (we don't run gulp in prod).

nickoneill:
@oderby once we start serving real urls (not hashes) this makes reloading easier while developing.

gulpfile.js
```diff
@@ -41,9 +42,17 @@ gulp.task('html:watch', function() {
 });
 
 gulp.task('html:serve', function (cb) {
+
+  function alwaysServeIndex(req, res, next) {
+    if(!(path.extname(req.url))) {
```
oderby, line 47: Please add a comment here on why this works/is the right way to do this (it took me a bit to realize that because the SPA routes are never for files, we can check for extensions on the request to see if it's a SPA route or an actual file request)

oderby (changes requested):
Now I see why this is helpful ;) Please rebase/merge and add the requested comment, then I think we should merge this.

gulpfile.js
greggraf: @oderby - created a new branch off the latest master and a new pull request here:  #270

I hope the code comment make sense. 
https://github.com/5calls/5calls/pull/270/files#diff-b9e12334e9eafd8341a6107dd98510c9R48

PR #268, opened 2017-03-28
Adds sending `via` parameter with call report

schraj:
Sends the text "web" for a call report post originating from the "5calls.org" website and the text "test" for a call report post originating from anywhere else.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -425,7 +425,10 @@ app.model({
 
       send('setUserStats', data, done);
       
+      // This parameter will indicate to the backend api where this call report came from
+      // A value of test indicates that it did not come the production environment
       const viaParameter = window.location.host === "5calls.org" ? 'web' : 'test';
+      
       const body = queryString.stringify({ location: state.zip, result: data.result, contactid: data.contactid, issueid: data.issueid, via: viaParameter })
       http.post(appURL+'/report', { body: body, headers: {"Content-Type": "application/x-www-form-urlencoded"} }, () => {
         // don’t really care about the result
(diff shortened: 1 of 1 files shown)


oderby:
Any way to test the production check locally? I've tried setting 5calls.org to redirect to 127.0.0.1 in my /etc/hosts file, but can't seem to get chrome or firefox to respect that change. I've verified it sends 'test' when run in normal development, and was hoping to test it by spoofing a prod environment.

If it's not as easy as I was hoping, then I'm still pretty comfortable saying this will work on prod...

schraj:
Yeah, I didn't find a way to test it locally.  I just verified that 

`window.location.host === "5calls.org"` returned `true`

when I stopped at a breakpoint on the production site at that point in the code.  

Not a satisfying test, definitely.

oderby (approved):

oderby:
Okay, well I guess we'll just have to confirm this works once we deploy next. FYI @nickoneill

schraj:
@oderby @nickoneill 

I was able to get "5calls.org" to map to "localhost:8000" by using Fiddler.  See second answer in this post:
http://stackoverflow.com/questions/8652948/using-port-number-in-windows-host-file

The problem then becomes that I'm running against the production backend which is also at 5calls.org so the app can't reach that.  

However, by putting the code that we're interested in:

window.location.host === "5calls.org" returns true

at the top of main.js and setting a breakpoint there I can verify that it works when running from "5calls.org" by setting the "via" parameter to "web".

Still not a great test, but closer.

Issue #267, opened 2017-03-27
Tests don't run when running them from Windows


schraj:
Karma works fine when I run it directly in this project and on other projects on my machine.

The issue is the "Spawn" process.  node can't find the karma package.  This is the error:

`Error: spawn ./node_modules/.bin/karma ENOENT
    at exports._errnoException (util.js:1022:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)`

I've played with the path some and not been able to fix it this way.

As a workaround, I've created another gulp task calls runs karma directly:

`gulp.task('test:js-unit-windows', function() {
    new Server({configFile: path.join(__dirname, "/karma.conf.js")}).start();
});`

I'm just using this workaround locally and haven't checked it in.  However, this is not a great solution because leaves open the chance for it to be inadvertently added at some point.

greggraf:
I was looking at this and it got me thinking about calling karma directly from npm scripts.  This might help with the spawn/windows problem. 

I have a branch here: https://github.com/greggraf/5calls/tree/move-karma-from-gulp-to-npm-scripts 

I don't have windows to test it, but if it fixes the problem, it might be worth discussing this approach. It simplifies the gulpFile.

(To compare to master: https://github.com/5calls/5calls/pull/271)

oderby:
I think I'm opposed to moving task/script definitions out of gulp and into package.json. I'm fine with everything being defined in gulp and mirrored/wrapped in npm scripts in package.json. But this approach leads to having  tasks/scripts defined in different locations, and seems like a recipe for pain down the road.

Would value @Mr0grog and @nickoneill opinions here, if they know of a better way to solve the problem of running tests on windows.

Mr0grog:
> I think I'm opposed to moving task/script definitions out of gulp and into package.json. I'm fine with everything being defined in gulp and mirrored/wrapped in npm scripts in package.json. But this approach leads to having tasks/scripts defined in different locations, and seems like a recipe for pain down the road.

I absolutely agree with this.

> The issue is the "Spawn" process. node can't find the karma package. This is the error:
> 
>     Error: spawn ./node_modules/.bin/karma ENOENT at exports._errnoException (util.js:1022:11) at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)

This _should_ be as simple as not being a doofus like me and using `path.join` to compose the path to Karma like so (though I could me missing something, I’ll need to spin up a Windows VM to check):

```js
[...]

PR #271, opened 2017-03-28
Move karma from gulp to npm scripts

greggraf:


schraj:
I've verified that this fixes #267 and that it now works on windows.  Thanks @greggraf!

greggraf:
While I totally understand the opposition to introducing a second place for defining task/scripts expressed in the comment of #267, I could not resist the temptation of 'simplifying' the code to 49 fewer lines.

PR #272, opened 2017-03-30
Attempt to fix problem running unit tests on Windows

greggraf:
- use path.join() to generate cross platform path to karma binary
- add shell:true option to help spawn command find the .cmd version of karma on windows


schraj:
@greggraf Yes, that works.  thanks for working that out.

oderby (approved):

PR #273, opened 2017-03-30
Localization

schraj:
This will have the localization implementation.

I need to build it up incrementally to see where the CI build is failing.  The previous build began to fail without providing much information and I was unable to figure it out from within that request.


schraj:
@Mr0grog Hi Rob, I've been struggling with the build for this pull request. @oderby indicated that you were familiar with this build and CircleCI more generally.

I think I've narrowed down the issue with these CircleCi failures on this pullrequest.  It appears to be some sort of timeout or usage throttling on our account based on the number of unit tests being run.  It passes until I have incrementally added unit tests up to a total count of about 60.  I tried deleting existing unit tests, such as the callcount_tests, and then I can add more of my new translation tests and it will pass as long as the number stays below that threshold.  The failure message is not helpful.  It just says that the VMs were "disconnected" and only 3 of them do this, the other VMs pass.  Does any of this trigger any ideas relative to the CircleCI account limits?

schraj:
After some chatting with @oderby, it appears it is a saucelabs issue as he can make it happen running gulp test:ci locally.

oderby:
I think I figured it out - code coverage seems to break edge and safari on sauce. See #274. With that change, one of the above [failing commits](https://github.com/5calls/5calls/pull/273/commits/0bb7afb01edd8b1d363a8410ee0ea7c84aa9eb4f) now pass for me.

I still have no idea why it worked reliably for <=60 tests...so maybe there's something else going on.

schraj:
This request is a duplicate now that #251 is building correctly.

Mr0grog:
Gah, sorry for not getting to this earlier. One thing you might want to try is setting the `concurrency` property in Karma’s configuration. I think we only get five concurrent runs on Sauce, so we might want to set this to `2` or `3` (so two PRs can run tests at the same time).

The more tests we add, the longer each browser run will take, and the more likely a browser will time out while waiting for Sauce to run it (because of its concurrency limits). The Karma `concurrency` property will prevent Karma from trying to start another browser run until others are done in that case, so we won’t have timeouts.

oderby:
@Mr0grog yeah, `concurrency` was the first thing I looked at. But in this case, it was unrelated. It's probably something we should look into setting soon.

As far what value to set it to - I would have thought CircleCI would have serialized PR tests, for exactly this reason - to not overload test infrastructure (SauceLabs, CircleCI's own vms, etc.). Is there documentation of that somewhere?

Mr0grog:
> I would have thought CircleCI would have serialized PR tests, for exactly this reason - to not overload test infrastructure

So, I know Circle can automatically split up tests for some kinds of test frameworks, but I think it can’t do that for us because we are running everything through Gulp (and I don’t know whether it has automatic support for Karma to begin with). We also have parallelism within single builds turned off, anyway (see https://circleci.com/gh/5calls/5calls/edit#parallel-builds). I haven’t done a lot with parallelism in Circle, though, so I’m not certain on the ins and outs.

The docs for how to manually set up parallelism are here: https://circleci.com/docs/1.0/parallel-manual-setup/, though I think there are significant changes for 2.0 (still in beta). Not sure on the details.

Since we have parallelism within a build turned off, I’m pretty sure we get those resources put towards multiple builds/PRs at once. Either way, Circle doesn’t have any smarts about how we are accessing external services like Sauce (doubly so since we are preventing Circle from seeing how we are using Karma by running it through Gulp), so we still have to make sure we never overrun Sauce when Circle is doing more than on thing at a time, whether that’s multiple builds or parallelism within a build.

Given that we can have 5 concurrent browser VMs in Sauce and 4 concurrent builds in Circle, the safest value for Karma’s `concurrency` setting would probably be `1`. I don’t think this project is supremely high-traffic on pushes to PRs, though, which is why I think we could get away with `2` or `3` and rarely wind up with one PR failing because it simply timed out before it could get test resources (we could also increase the `captureTimeout`, too, though it’s already really high).

Mr0grog:
Sorry, that was a long-winded way to say:

- They are serializing tests within a single PR (but that does not serialize how Karma utilizes browsers on Sauce within the tests for a single PR, and in CI, Karma is trying to test on 8 browsers, which is more than I’m guessing our plan on Sauce allows for at once)
- They are parallelizing (I think; have not tested) the latest test for each PR—so if two people push to two different PRs at the same time, both those PRs will be tested simultaneously.

PR #274, opened 2017-03-31
Disable code coverage for CI tests.

oderby:
For some reason, if we run code coverage,
edge and safari browsers hang on Sauce.

Disabling it for now until the issue is
resolved upstream. See
https://github.com/istanbuljs/babel-plugin-istanbul/issues/81

### DIFF
--- a/karma.ci.conf.js
+++ b/karma.ci.conf.js
@@ -109,6 +109,16 @@ module.exports = function (configuration) {
     reporters: ['mocha', 'saucelabs'],
     singleRun: true,
 
+    // don't apply code coverage transformation, it breaks things on sauce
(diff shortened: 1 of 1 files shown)


oderby:
@greggraf fyi, this appears to fix the troubles @schraj was seeing with saucelabs tests in #251. I'll wait for him to confirm.

schraj:
That fixes it! Nice work @oderby.  Thanks! #251 now passes with my full suite of tests and changes.

greggraf (approved):
Wow, Great find. For the future, I would suggest we probably only want to generate one set of coverage anyway - instead of one for every browser on SauceLabs.

PR #275, opened 2017-04-01
Example sitemap with gulp support

nickoneill:
This isn't a dynamically generated sitemap, it's an example with the current issues. I realized that google is *still* not finding our issue pages because it doesn't wait long enough for the issues list to load. So we provide a sitemap for all available issues instead, for now.


nickoneill:
nm actually, I can test this manually for now and this can be progress for making this build based off live issues list

Issue #180, opened 2017-02-11
Link to town halls


remoun:
Link to (or better yet, embed) info about upcoming town halls, by location.

For example, the [Town Hall Project](https://twitter.com/townhallproject) has compiled a [spreadsheet](https://docs.google.com/spreadsheets/d/1yq1NT9DZ2z3B8ixhid894e77u9rN5XIgOwWtTW72IYA/htmlview?sle=true#gid=1473996386)

nickoneill:
Thanks @remoun, we're working on this.

nickoneill:
Fixed in #286

PR #286, opened 2017-04-11
Link to town hall data

murribu:
This is an attempt to loop in data from [The Town Hall Project](https://townhallproject.com/).

https://github.com/5calls/5calls/issues/180


nickoneill:
Thanks @murribu, I'll check this out shortly. FYI for anyone else testing, requires some district changes on the backend that are in PR.

static/js/components/townHall.js
```diff
@@ -16,7 +16,7 @@ let renderNotes = (e) => {
 };
 
 module.exports = (state) => {
-  if (state.localEvents.length > 0){
+  if (!!state.localEvents && state.localEvents.length > 0){
```
oderby: double negation isn't needed here - value is already cast to boolean when evaluated as part of conditional.

static/js/main.js
```diff
@@ -7,11 +7,15 @@ const logger = require('loglevel');
...
 const store = require('./utils/localstorage.js');
 const scrollIntoView = require('./utils/scrollIntoView.js');
+const townHallUtils = require('./utils/townHallUtils.js');
+const zipcodes = require('zipcodes');
 
 const app = choo();
 const appURL = 'https://5calls.org';
 // const appURL = 'http://localhost:8090';
 
+const maxTownHallDistance = 50;
```
oderby: please ad a comment on this value. What are the units? Why was 50 chosen? When should it be changed?

static/js/main.js
```diff
@@ -241,11 +249,54 @@ app.model({
...
+    setCacheDate: (state, data) => ({ [data]: Date.now() }),
+    receiveTownHallData: (state, data) => {
+      let events = JSON.parse(data);
+      let lat = false;
+      let lng = false;
+      let locationExists = !!localStorage['org.5calls.location'];
+      let geolocationExists = !!localStorage['org.5calls.geolocation'];
+      if (locationExists){
+        // zipcode
+        let zipcode = localStorage['org.5calls.location'].replace(new RegExp(/\]|\[|"/, 'g'),'');
```
oderby: Please use (or extend/fix, if needed) our localStorage utility module, ./utils/localstorage.js, for interacting with localstorage.

static/js/main.js
```diff
@@ -241,11 +249,54 @@ app.model({
...
+          lng = zipcode_obj.longitude;
+        }
+      } else if (geolocationExists){
+        // lat/long
+        let geo = localStorage['org.5calls.geolocation'].replace(new RegExp(/\]|\[|"/, 'g'),'');
+        lat = geo.split(",")[0];
+        lng = geo.split(",")[1];
+      }
+      if (!!lat && !!lng){
+        events = Object.keys(events).map(function(key){ return events[key]; }); // Convert Object to Array
```
oderby: Why not just use `Object.values` method?

static/js/main.js
```diff
@@ -241,11 +249,54 @@ app.model({
...
+        events = events
+          .map(townHallUtils.mapDistance(lat, lng))
+          .filter(townHallUtils.filterEvents(state, maxTownHallDistance))
+          .sort(townHallUtils.sortEvents()); // Sort by distance
+        return {
+          localEvents: events.slice(0,3)
```
oderby: Why only closest 3? Might be worth a comment.

static/js/main.js
```diff
@@ -241,11 +249,54 @@ app.model({
...
+      }
+    },
   },
 
   effects: {
+    fetchTownHallData: (state, data, send, done) => {
+      let townHallUrl = "https://townhallproject-86312.firebaseio.com/townHalls.json";
+      logger.debug("fetching url", townHallUrl);
+      http(townHallUrl, (err, res, body) => {
+        send('receiveTownHallData', body, done);
```
oderby: Please add error handling.

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+  "UT": "Utah",
+  "VT": "Vermont",
+  "VI": "Virgin Islands",
+  "VA": "Virginia",
+  "WA": "Washington",
+  "WV": "West Virginia",
+  "WI": "Wisconsin",
+  "WY": "Wyoming",
+}
+let calculateDistance = (lat1, lon1, lat2, lon2) => {
```
oderby: Can you link to a source or give an explanation of the math here in a comment?

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+  let radlat2 = Math.PI * lat2/180
+  let theta = lon1-lon2
+  let radtheta = Math.PI * theta/180
+  let dist = Math.sin(radlat1) * Math.sin(radlat2) + Math.cos(radlat1) * Math.cos(radlat2) * Math.cos(radtheta);
+  dist = Math.acos(dist)
+  dist = dist * 180/Math.PI
+  dist = dist * 60 * 1.1515
+  return dist
+};
+// calculate the distance between two lat/long points
```
oderby: comment probably belongs on above function?

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+module.exports = {
+  calculateDistance: calculateDistance,
+  parseCivicData: (divisions) => {
```
oderby: Either define all methods in object or separately, as you do above. Don't mix styles.

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+    return ret;
+  },
+  stateAbbrs: stateAbbrs,
+  mapDistance: (lat, lng) => {
+    return (e) => {
+      e.distance = calculateDistance(lat, lng, parseFloat(e.lat), parseFloat(e.lng));
+      return e;
+    };
+  },
+  filterEvents: (state, maxTownHallDistance) => {
```
oderby: This logic is complex enough that it warrants some unit tests.

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+  let theta = lon1-lon2
+  let radtheta = Math.PI * theta/180
+  let dist = Math.sin(radlat1) * Math.sin(radlat2) + Math.cos(radlat1) * Math.cos(radlat2) * Math.cos(radtheta);
+  dist = Math.acos(dist)
+  dist = dist * 180/Math.PI
+  dist = dist * 60 * 1.1515
+  return dist
+};
+// calculate the distance between two lat/long points
+module.exports = {
```
oderby: Please add unit tests for these methods, especially for testing around corner cases and error cases.

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+        'state': false,
+        'stateAbbr': false,
+    };
+    
+    for(let i in divisions){
+      let parts = divisions[i].split('/');
+      if (parts.length > 0){
+        let key = parts[parts.length - 1].split(':')[0];
+        let val = parts[parts.length - 1].split(':')[1];
+        switch (key){
```
oderby: Rather than relying on all 3 types of divisions to be returned/provided, wouldn't it be more robust to extract this just from the congressional district division? That is, the division specifying the user's congressional district, which looks like `ocd-division/country:us/state:fl/cd:18`, has all the information you need to extract cd, state and country. It seems more robust to just use that.

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+  mapDistance: (lat, lng) => {
+    return (e) => {
+      e.distance = calculateDistance(lat, lng, parseFloat(e.lat), parseFloat(e.lng));
+      return e;
+    };
+  },
+  filterEvents: (state, maxTownHallDistance) => {
+    return (e) => {
+      return e.dateObj > (new Date()).getTime() // The event is in the future
+            && ((!!state.divisions.congressional_district
```
oderby: as far as I can tell, you only need access to the `divisions` object on the state object. Can you pass that into the method, rather than the entire state object? I'm asking mostly because in this context, I think state refers to a state in the USA, and find it confusing that it's not. But also because it's clearer to pass in just the minimal state needed. If you do decide you need to continue passing in the full state object, please clarify the variable name.

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,128 @@
...
+  },
+  filterEvents: (state, maxTownHallDistance) => {
+    return (e) => {
+      return e.dateObj > (new Date()).getTime() // The event is in the future
+            && ((!!state.divisions.congressional_district 
+                && state.divisions.congressional_district == e.District) // The event is for the user's District
+              || (!!state.divisions.state 
+                && state.divisions.state == e.State 
+                && e.District == "Senate" 
+                && e.distance < maxTownHallDistance)) // The event is for a Senator in the user's state, less than maxTownHallDistance away
```
oderby: Since this is the only place this constant is used, perhaps we can just define it within this module and keep it entirely contained within the module internals (ie don't expose it to the rest of the app)?

static/js/main.js
```diff
@@ -241,11 +249,54 @@ app.model({
...
+        return {
+          localEvents: []
+        };
+      }
+    },
   },
 
   effects: {
+    fetchTownHallData: (state, data, send, done) => {
+      let townHallUrl = "https://townhallproject-86312.firebaseio.com/townHalls.json";
```
oderby: Can you add a brief comment on who provided this url/where it came from?

static/js/main.js
```diff
@@ -241,11 +249,54 @@ app.model({
...
+        let geo = localStorage['org.5calls.geolocation'].replace(new RegExp(/\]|\[|"/, 'g'),'');
+        lat = geo.split(",")[0];
+        lng = geo.split(",")[1];
+      }
+      if (!!lat && !!lng){
+        events = Object.keys(events).map(function(key){ return events[key]; }); // Convert Object to Array
+        events = events
+          .map(townHallUtils.mapDistance(lat, lng))
+          .filter(townHallUtils.filterEvents(state, maxTownHallDistance))
+          .sort(townHallUtils.sortEvents()); // Sort by distance
```
oderby: I feel like this is a bit of a leaky abstraction, having the townHallUtils module export these 3 methods which are sort of independent, but they'll only work correctly if used in this order. Perhaps it'll be simpler to move them internal to the module and expose just a single method `filterForLocalEvents` or such? (you might still want to expose the individual methods for unit testing, which is fine)

static/js/main.js
```diff
@@ -241,11 +249,54 @@ app.model({
...
+          lat = zipcode_obj.latitude;
+          lng = zipcode_obj.longitude;
+        }
+      } else if (geolocationExists){
+        // lat/long
+        let geo = localStorage['org.5calls.geolocation'].replace(new RegExp(/\]|\[|"/, 'g'),'');
+        lat = geo.split(",")[0];
+        lng = geo.split(",")[1];
+      }
+      if (!!lat && !!lng){
```
oderby: `lat` and `lng` are already treated as boolean values by the conditional, so no value gained from the double negation here.

static/js/components/townHall_test.js
```diff
@@ -0,0 +1,72 @@
...
+        timeZone:"EDT",
+        yearMonthDay:"2017-04-04"
+      }];
+describe('townHall component', () => {
+  describe('eventList include', () => {
+    it('should display a list of events when found', () => {
+      let state = {localEvents: localEvents};
+      let result = townHall(state);
+      let ul = result.querySelector('.town-hall ul');
+      expect(ul).to.be.defined;
```
oderby: Can you add an assertion on the number of events?

oderby (commented):
Some drive-by comments on the code. Haven't tried running it yet. @nickoneill sorry for extra review if you're already looking at it...

nickoneill:
This looks incredibly good @murribu! A couple design notes:

* I love the event type badge in the upper right, great way to tell what an event is
* Let's keep the "learn why calling" text above the event list, immediately following the share links. Moving the total calls below events is fine.
* Town hall attribution can just be "Events from the Town Hall Project", sans extra `*`
* Some events with longer descriptions can feel like a wall of text. Worth truncating to 4-5 lines and expecting users to follow the link if they want to read me?
* Break the title text into distinct sections by bolding the event title: "Town Hall Event - With or Without..." in the included screenshot.

image

static/js/utils/townHallUtils.js
murribu: I found that some zipcodes would return the user's state, but not their congressional district. My guess is that the district line passes through the zipcode, so the civic api can't specify one district. In that case, I want to still have access to the state, even though I don't have the congressional district.

static/js/main.js
murribu: @nickoneill provided this url for me. I'll let him comment on its source

murribu:
I believe I've addressed most of the requested changes.
Notably, I did not truncate any events' descriptions - because some of them are html with a link to More Info at the end. I suppose we would need to parse the html and pull out the link if we want to truncate those descriptions.
Also, I did not add a comment for the THP json url. I'd like @nickoneill to provide its source info.

static/js/main.js
nickoneill: Townhall project gave us this URL so we could integrate with their service.

nickoneill: On second thought, I'm guessing @oderby is asking us to actually put the source as a comment in here to record for the future. "Data provided from the Town Hall Project, http://townhallproject.com" is probably good

nickoneill (approved):

nickoneill:
Perfect, waiting for circle to complete and we'll bring this in.

nickoneill:
I'm actually going to give i18n a day to wrap up and merge before this, I'd like to get that base work in before making more merge conflicts for that PR. This one should come in cleanly following it.

oderby:
Now that i18n is in, this will need to be rebased and all strings will need to be localized. I guess for now we're just using google translate, and having them checked in #i18n channel. You can see in #251 example of how the overall i18n system works. @nickoneill we may want to come up with a more streamlined way to get translations for PRs with new content...

nickoneill:
OK, I'll put together a quick doc with the needed strings tomorrow and we can see who's got time to translate on slack.

nickoneill:
Whoa, I totally missed that this wasn't in yet. Let's fix that.

Failures are all eslint errors, I can't push to this branch but `eslint --fix` on the failed files should do it. Working on translation strings now.

nickoneill:
Made some eslint changes and added translation strings (in `en`) here: https://github.com/5calls/5calls/tree/murribu-murribu-town-hall-data

murribu:
Oops. I just saw that you created a different branch for this. Sorry about that.

nickoneill:
Not all, I wanted to check in some of the basic work I added in but we should pull those commits into here.

nickoneill:
@murribu we can either re-PR the other branch or you can cherry-pick those changes into this branch. Would prefer to keep things here rather than duplicate but let me know if you don't have time and I'll merge the other one.

murribu:
I should have some time to look at this Monday

nickoneill:
👍

nickoneill:
Test failure was just a funky safari vm thing. Works now! Thanks!

PR #289, opened 2017-04-13
Clearer Location Form

nickoneill:
Reverts a revert. We'll fix the e2e tests for location here after they're merged in.

I'm dumb and we are planning on merging the e2e tests in #252 before this. We'll do those first, then circle back on fixing the e2e tests for the changed location button.

**Don't merge me yet**


nlsandler:
I messed up the commit history here, so I've opened #292 instead.

PR #292, opened 2017-04-15
Clearer Location Form

nlsandler:
Reverts "Revert Clearer Location Form" and updates the e2e tests for those changes.

This should replace #289 because I screwed up the history in that one.


nickoneill (approved):
Tests look great! Thanks for adding those in

PR #299, opened 2017-04-21
Adds dynamic scripts

bernardwang:
## Summary
Adds new component `scriptFormat.js` to replace specified script placeholders. It uses regex to replace the script, and currently will replace location and representative titles. The PR also removes a check in `setCachedCity()` so that `state.cachedCity` is updated more frequently. 

`[CITY, STATE]` => Urbana
`[REP/SEN NAME]` => Senator Duckworth
`[REP/SEN NAME]` => Rep. Davis

## Tests
Manual testing and unit tests were added.
Before:
Before
After:
After

## Notes
Currently only the city is provided in `response.normalizedLocation`, when ideally State or ZIP would also be useful. An option could be to modify the backend to also return State or ZIP from Google Civic API. 

For the future, other dynamic fields could be bill names like "[HR 1880 for House/S 806 for Senate]", but this would require more standardization on copy's end.

### DIFF
--- a/static/js/components/script.js
+++ b/static/js/components/script.js
@@ -2,7 +2,7 @@ const html = require('choo/html');
 const t = require('../utils/translation');
 
 const find = require('lodash/find');
-const scriptLine = require('./scriptLine.js');
+const scriptFormat = require('./scriptFormat.js');
 
 module.exports = (state, prev, send) => {
   const issue = find(state.issues, ['id', state.location.params.issueid]);
@@ -13,7 +13,7 @@ module.exports = (state, prev, send) => {
     return html`
       <div class="call__script">
         <h3 class="call__script__header">${t("script.yourScript")}</h3>
-        <div class="call__script__body">${issue.script.split('\n').map((line) => scriptLine(line, state, prev, send))}</div>
+        ${scriptFormat(issue, state, prev, send)}
       </div>`;      
   } else {
     return html``;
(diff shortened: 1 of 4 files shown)


cdoremus:
@bernardwang  is this code able to handle the case where the last name is two or more words, which is the case for our 8th president (Martin Van Buren) and former California state senator Antonio Maria de la Guerra (Spanish surnames can go on forever)? I noticed that our Airtable test database only has one field for the name (not sure about the prod version).

Thinking more about this, we also might want to remove last-name suffixes such as Jr, Sr, II, III, or IV. That might be a little easier to parse out since there appears to be a small set of possible suffixes.

nickoneill:
@cdoremus Prod airtable is pretty similar to development (dev has a new category field I'm working on, otherwise the same). Most contacts are coming from Google Civic but they're the same: a single field as a name.

Sounds needlessly complex to split out last name given the limitations we have for name format and the various styles associated. We could easily just say "Senator Cory Booker" or "Rep. Antonio Maria de la Guerra" for now and improve for a v2 API.

nickoneill:
Sorry for the wait here @bernardwang, will review today!

nickoneill (approved):

nickoneill:
This looks great @bernardwang! It makes the scripts way more natural. I'll pull this in now and start instructing the issues team about how to revise the scripts to utilize this.

Issue #293, opened 2017-04-20
Hide impact in footer if there's no user stats yet


nickoneill:
There's a link in the footer component that shows your call counts so far. It should hide if there aren't any user stats for calls yet.

nickoneill:
Completed in #300

PR #300, opened 2017-04-22
Rendering the impact link in the footer

greggraf:
To address #293, render the impact link only if the user has stats available.

Checks whether the array `userStats.all` exists and has length.


nickoneill (approved):

nickoneill:
Great, thanks!

PR #301, opened 2017-04-22
disable <span> tag in alt attribute for images

greggraf:
@schraj - I found a few spots where the translation function needed to have the `justText` flag passed in the arguments.


nickoneill:
Interesting - just discovered these out of chance?

nickoneill (approved):

greggraf:
@nickoneill - yes, I noticed it because I had my local app in a situation where images were not in the right path and the broken image alt text looked wrong.

schraj:
@greggraf nice catch : )

PR #305, opened 2017-05-03
removed logic to add +1 to phone numbers 

greggraf:
in expectation that the server will provide. Updated logic in href to remove space in phone numbers

### DIFF
--- a/static/js/components/contact.js
+++ b/static/js/components/contact.js
@@ -24,7 +24,7 @@ module.exports = (c, state, prev, send) => {
           <h3 class="call__contact__field-offices__header">${t('contact.localOfficeNumbers')}</h3>
           <ul class="call__contact__field-office-list">
             ${c.field_offices.map(office => html`
-              <li><a href="tel:+1${office.phone.replace(/-/g, '')}">${office.phone}</a> ${cityFormat(office,c)}</li>
+              <li><a href="tel:${office.phone.replace(/-| /g, '')}">${office.phone}</a> ${cityFormat(office,c)}</li>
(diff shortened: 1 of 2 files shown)


nickoneill:
Thanks! This'll make it easy to roll this out when we change the backend.

Test fails on a js error, fix looks like this: https://github.com/5calls/5calls/tree/pr/305

static/js/components/contact.js
```diff
@@ -24,7 +24,7 @@ module.exports = (c, state, prev, send) => {
           <h3 class="call__contact__field-offices__header">${t('contact.localOfficeNumbers')}</h3>
           <ul class="call__contact__field-office-list">
             ${c.field_offices.map(office => html`
-              <li><a href="tel:+1${office.phone.replace(/-/g, '')}">${office.phone}</a> ${cityFormat(office,c)}</li>
+              <li><a href="tel:${office.phone.replace(/-| /g, '')}">${office.phone}</a> ${cityFormat(office,c)}</li>
```
oderby: Visually, it looks cluttered/clumsy to *display* the phone number with the "+1 " prefix, so I think we should remove it. Perhaps we can strip the prefix for the display only? Something like `office.phone.replace(/^\+\d* /, '')` here and below.

greggraf:
@oderby - I updated the tests to verify the formatting of both the href and the display text and changed the code with what I think you were suggesting.

oderby (approved):
lgtm. Tested that this works against current backend, so I think this is safe to merge as is, independent of backend changes.

Issue #303, opened 2017-04-23
Support switching languages in UI


nickoneill:
Now that we support spanish in the UI, we should provide a toggle to manually set the language rather than exclusively relying on browser language which might not accurately represent the language desired.

oderby:
Per conversation in slack, it would be nice to report language choice statistics to our Google Analytics account, so we can better understand the language breakdown of our users. Here are the directions to do so via @nickoneill 

> I just set up a custom dimension 
>```javascript
>var dimensionValue = 'SOME_DIMENSION_VALUE';
> ga('set', 'dimension3', dimensionValue);
>```
> Keep `dimension3`, set the value to something to uniquely identify the language

@greggraf I think it'd make sense to include this in #312, but wanted to get your thoughts before increasing the scope of that work.

greggraf:
It is pretty easy to set that when the language is changed.
And when the app starts up, it can be set to the default or what was pulled from localStorage.

Both of those occur after the Google Analytics library has loaded and the first calls from the html page have already been queued up and sent off.

This means the initial calls will *not* have the 'dimension3' set, but further reporting calls from within the app will (such as navigating to issue pages or indicating the results of calls).

It is a little trickier to get that information into the initial calls from the html page. The app does the work to extract the localStorage value or set the default, but the html page does not have this information when its loading the Google Analytics library and queuing up the reporting calls.

I will put together a commit that makes the post app loading reporting calls have this value and if its worth pursing I will explore getting that value out to the other calls.

greggraf:
ok, i have the following approach:

https://github.com/5calls/5calls/pull/312/files#diff-93471d4bf84d0569fadcebe5fba72773R89
In the initial execution of `main.js` after determining the `cachedUserLocale` from either `localStorage` or the `i18n.js` library, store that value in the `userLocale` property on the `window` object.

https://github.com/5calls/5calls/pull/312/files#diff-0a3e976c1c22ad5508a7403949f5faf9R115
In `index.html`, during the Google Analytics code, check if `userLocale` has been set on the `window` object, set it to the `dimension3` value after creating the Google Analytics instance, but before sending the pageview call.

The check for `userLocale` is there in case the code determining `cachedUserLocale` has not finished or fails before the Google Analytics sends the pageview call.

https://github.com/5calls/5calls/pull/312/files#diff-93471d4bf84d0569fadcebe5fba72773R262
[...]

PR #312, opened 2017-05-11
Support switching languages in UI

greggraf:
This is a first pass at https://github.com/5calls/5calls/issues/303 - Support switching languages in UI

Added a drop down in the footer which displays the configured available language codes.
Selecting a language code switches the translation strings to the selected language and stores the choice in local storage.


nickoneill:
Thanks! This looks great on web, I'll test on mobile as soon as I'm back tonight.

One semi-unrelated thing about code that popped into my head while reviewing: do we think localstorage strings such as `org.5calls.userlocale` should be stored in `constants.js` rather than relying on magic strings in various parts of the code?

greggraf:
So I started thinking about the constants question, and my first thought was "yes, those strings should be constants". My second thought was "but we only use these strings in one file, `main.js` so maybe we don't need to move them to `constants.js` until we see a need to share them across files.

Then I got distracted and started looking at ways to clean up the localStorage code.  I felt like i needed to have tests first, so I opened this PR: "Add unit tests for localStorage and default state" #316.

Then I created the following branch with more tests, and actual updates to the code:
   https://github.com/5calls/5calls/compare/master...greggraf:refactor-localstore?expand=1

Not a PR yet, because I am recognize that this kind of work may not be of interest at this time. 

output of tests for `static/js/utils/localstorage.js`
```
  localStorage utils
    getAll()
      ✔ should return an empty array on error 
      ✔ should return an empty array if not defined 
      ✔ should return the stored array if defined
    get()
      ✔ should return first item in the list 
      ✔ should return undefined if list is empty
      ✔ should return undefined if key does not exist
    set()
      ✔ should replace first item in the list 
    add()
      ✔ should add the new value to the end of the list
    replace()
      ✔ should replace the item at the given index with the new value
    remove()
      ✔ should call localStorage.removeItem api with key to be removed
```

static/js/main.js
```diff
@@ -137,6 +139,8 @@ app.model({
     showFieldOfficeNumbers: false,
 
     debug: debug,
+    constants: constants,
```
oderby: This feels wrong to me: if they're constants, they shouldn't change, so they don't belong in the state. If they do change, then they aren't really constants...

However, I see why you're doing this: accessing them this way allows you to inject custom/different values from your unit tests. Unfortunately, I can't think of any better way to do this. Perhaps we just rename them to be something like `globalConfig` or just `config` to avoid the apparent paradox? And also include a comment about why you're exposing them this way?

static/js/main.js
```diff
@@ -253,7 +257,13 @@ app.model({
     },
     toggleFieldOfficeNumbers: (state) => ({ showFieldOfficeNumbers: !state.showFieldOfficeNumbers }),
     hideFieldOfficeNumbers: () => ({ showFieldOfficeNumbers: false }),
-    setCacheDate: (state, data) => ({ [data]: Date.now() })
+    setCacheDate: (state, data) => ({ [data]: Date.now() }),
+
+    translate: (state, data) => {
```
oderby: nit: this is actually where the language is changed, so this should be called `changeLanguage`. Then perhaps rename the effect to `chooseLanguage`?

static/js/main.js
```diff
@@ -456,8 +466,18 @@ app.model({
       ga('send', 'event', 'issue_flow', 'select', 'select');
 
       scrollIntoView(document.querySelector('#content'));
+    },
+
+    changeLanguage: (state, data, send, done) => {
+      localization.change(data, function(err){
+        if (!err) {
```
oderby: When would this fail? Should we do anything to alert the user if there is an error?

static/js/utils/localization_test.js
```diff
@@ -32,4 +32,24 @@ describe('userLocaleDetection', () => {
       expect(result).to.equal(expected);
     });
   });
+
+  describe('change', () => {
+    it('should load the new language as passed', (done) => {
+
+      l.change('es', function(err, t) {
```
oderby, line 39: nit: always good practice to check that err is not set.

static/js/components/footer.js
```diff
@@ -17,6 +17,33 @@ module.exports = (state) => {
...
+
+      var options = locales.map(function( locale ) {
+
+        var selected = "";
+
+        if (locale === state.selectedLanguage) {
+          selected = "selected";
+        }
+
+        return html`<option ${selected} value="${locale}">${locale}</option>`;
```
oderby: I think it'd be more user friendly to use the full name for the language in the option text. That is, "English" and "Español", rather than "en" and "es".

That said, I'm unsure if those strings themselves should be localized or not...

oderby (commented):
Some minor comments. The only major question I have is if it makes sense to use a dropdown or just an href to click and toggle the state. Obviously that only works as long as we only support two languages, but that seems like a safe assumption to me. Thoughts?

static/js/main.js
greggraf: I was attaching constants to the state not only for the tests, but because generally state is used in the view js files for interpolation into the templates. Having seen the way constants are included and used in the `about.js` and `call.js` files, I have reworked the `footer.js` code to be similar.

It's probably unnecessary to inject these values into the tests. If the tests are written against known, manually maintained constant values, when those change, the test expected output will change.

greggraf:
Any thoughts here?

This is a screenshot of an alternative approach to the drop down. 
- It uses  "English" and "Español" (not translated). 
- This approach assumes there will only be the two options to toggle between.
- When the text is in English, "Español" is visible, when the test is in Spanish, "English" is visible

option

static/js/main.js
greggraf: The flow:
 - user UI interaction calls `send("changeLanguage")`
 - `changeLanguage()` calls `localization.change()` which calls `i18n.changeLanguage()` and attempts to translate the text to the selected language
 - if `i18n.changeLanguage()` does not throw an error it calls `send("translate")`
 - `translate()` updates `localStorage` and `state` with the new selectedLanguage
 
Changing `changeLanguage()` to `chooseLanguage()` makes lots of sense.

Perhaps `translate()` should become `languageChanged()` since it gets invoked after the vendor library method `i18n.changeLanguage()` has been successful?

static/js/main.js
greggraf: `i18n.changeLanguage()` will throw an error if something goes wrong trying to load the new translation file. It would be friendly to alert the user. We can throw up an old school alert box or - if there is already something in place for showing modal boxes even better. I will look around the code base.

static/js/main.js
oderby: `languageChanged` sounds good to me.

oderby:
That UI looks great to me! Much cleaner than a dropdown. And we can always revisit if we do decide to support more than 2 languages.

static/js/utils/localization_test.js
greggraf: is this the idea?

greggraf:
I think there is a problem with the sauce credentials:
```
03 06 2017 14:53:26.585:ERROR [launcher.sauce]: Can not start chrome latest (Windows 10)
  Failed to start Sauce Connect:
  Invalid Sauce Connect Credentials. 3 Jun 14:53:26 - Response: {"message": "Not authorized"}.
```

nickoneill:
Huh. I don't see anything immediately wrong in sauce, let's try a rebuild...

nickoneill:
Uh, nope. I'll send them an email.

static/html/index.html
```diff
@@ -112,6 +112,9 @@ <h2 class="callcount">5 Calls will be right back.</h2>
       })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
       ga('create', 'UA-90915119-1', 'auto');
+      if (window.userLangauge) {
```
oderby: shouldn't this be checking that `window.userLocale` exists?

static/html/index.html
```diff
@@ -112,6 +112,9 @@ <h2 class="callcount">5 Calls will be right back.</h2>
...
+        ga('set', 'dimension3', window.userLocale);
```
oderby: can we get a comment in the code on what this is doing - `'dimension3'` isn't very descriptive, unfortunately.

static/js/components/footer.js
```diff
@@ -1,46 +1,47 @@
...
     if (locales) {
-
-      var options = locales.map(function( locale ) {
-
-        var selected = "";
-
-        if (locale === state.selectedLanguage) {
-          selected = "selected";
+      return locales.map(function( locale ) {
+        if (locale !== state.selectedLanguage) {
```
oderby: nit: should probably also require that `locale` is in `displayLocale` here. So this should be 

```javascript
if (locale !== state.selectedLanguage && displayLocale[locale]) {
```

static/html/index.html
greggraf: yes! last minute change of variable without and IDE or unit tests.

static/js/components/footer.js
greggraf: yes!

greggraf:
@nickoneill I got an email from CircleCI yesterday... wondering if you got the same?

nickoneill:
@greggraf I'm going back and forth with support over email. I haven't received anything else, something useful?

nickoneill:
Ok, figured it out. Circle CI doesn't export the environmental variables for forks, because doing so would mean that anyone could just `echo $SECRET_KEY` and get any keys. You're a team member so you should be able to make non-fork PRs, but I went ahead and disabled it for a moment. Passing again!

oderby (approved):
needs rebase/merge, but otherwise lgtm

greggraf:
@oderby and @nickoneill - is it up to me to also close the issue? Or does the person who opened it do that?

oderby:
go ahead and close it :)

Issue #310, opened 2017-05-10
Expand dynamic scripts


bernardwang:
### Summary 
Currently scripts are being formated so that location and name of representatives are dynamically filled. This is implemented with regex to replace certain strings like `[STATE/ZIP]` and `[REP/SEN NAME]`. Considering recent issues have included more numbers (committees, attorney generals), this could be expanded to make scripts more natural.

Here is an example of a script that could be made dynamic:
![script](https://cloud.githubusercontent.com/assets/5424713/25882470/3c9ce574-3509-11e7-955a-8bbb0cdcc333.png)

Specifically, these are common cases that could be handled
`[Senate: S. 100; House: H.R. 100]` - Dynamically show bill names in senate and house
`[IF CALLING HOUSE/SENATE: ...]` - Showing for representatives
`[IF CALLING REPS, ADD: ...]`- Show for House
`[IF CALLING SENATOR, ADD: ...]` - Show for Senate
`[IF CALLING COMMITTEE, ADD: ...]` - Show for committee

If the current implementation is continued, these cases should be standardized to keep the regexes simple.

piebob:
love this. i am about to file an issue (i'll link it to this) requesting that we also make the "why you're calling this office" content dynamically settable.

piebob:
i guess we can't link stuff in github issues :)  but it's this one: https://github.com/5calls/5calls/issues/311

nickoneill:
From a PR that didn't quite make it into the old choo branch:

> Hey @bernardwang, sorry for the very extended wait on this. I've been thinking about how to approach this recently, talking to users who like our customized scripts and thinking about how the policy team makes scripts.

> Making vast improvements here would definitely be valuable. But what if we want to do more detailed things like if/else blocks? Sometimes we call two senators and a committee and the scripts are subtlety different for each one, but adding these subtleties gives people the confidence to call.

> Does it make sense for us to expand this script templating language and implement it on each platform? Or make a roundtrip to the server asking for a script for each one? Or just include each script with a contact in the original issues request? Pros and cons for each, genuinely wondering what you think on the architecture here.

PR #313, opened 2017-05-13
Expand dynamic scripts

bernardwang:
## Summary
Issue #310 

This PR refactors `scriptFormat.js` and adds 3 cases to dynamic scripts.
1. Senators/Reps/Attorney General names - currently only `[REP/SEN NAME]` is being targeted, this adds support for replacing `[REP NAME]`, `[SEN NAME]`, `[ATTORNEY GENERAL NAME]`.
2. Bill Names - displays only the correct bill title in clauses like `[House- H.R. 1180, Senate: S 801]`. Currently requires two regexs because copy is not standardized to easily match this case.
3. Committee - conditionally displays scripts for committees `[IF COMMITTEE, ADD: ...]`. Same idea can be extended to `[IF SENATE: ...]` and other conditional sentences. Also requires standardization in copy.

I think adding more cases will make the scripts more natural and content aware. However, standardizing copy is important in making the regexes simpler and easier to test.  Also, the number of regex replacements could potentially get quite large, but I am attempting to limit them by only applying the expressions relevant to the current contact. I am open to any suggestions on how this could be better implemented or standardized! 

## Examples
[Case 1 Before](https://cloud.githubusercontent.com/assets/5424713/26022673/74951032-3770-11e7-9192-51028912b71e.png)
[Case 2 After](https://cloud.githubusercontent.com/assets/5424713/26022674/77b421a4-3770-11e7-9bb3-4f3d7f3a7642.png)

[Case 2 & 3 Before](https://cloud.githubusercontent.com/assets/5424713/26022683/b109d5a2-3770-11e7-8530-35aa2985ce18.png)
[Case 2 & 3 After](https://cloud.githubusercontent.com/assets/5424713/26022684/b3bfa4f2-3770-11e7-99d4-e51f7715059c.png)

### DIFF
--- a/static/js/components/script.js
+++ b/static/js/components/script.js
@@ -8,7 +8,7 @@ module.exports = (state, prev, send) => {
   const issue = find(state.issues, ['id', state.location.params.issueid]);
   const currentIndex = state.contactIndices[issue.id];
   const currentContact = issue.contacts[currentIndex];
-    
+
   if (currentContact != null) {
     return html`
       <div class="call__script">
--- a/static/js/components/scriptFormat.js
+++ b/static/js/components/scriptFormat.js
@@ -4,27 +4,66 @@ const scriptLine = require('./scriptLine.js');
 module.exports = (issue, state) => {
   const currentIndex = state.contactIndices[issue.id];
   const currentContact = issue.contacts[currentIndex];
+  let script = issue.script;
 
   // Replacement regexes, ideally standardize copy to avoid complex regexs
-  const titleReg = /\[REP\/SEN NAME\]|\[SENATOR\/REP NAME\]/gi;
-  const locationReg = /\[CITY,\s?ZIP\]|\[CITY,\s?STATE\]/gi;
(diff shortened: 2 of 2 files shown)


nickoneill:
Hey @bernardwang, sorry for the *very* extended wait on this. I've been thinking about how to approach this recently, talking to users who like our customized scripts and thinking about how the policy team makes scripts.

Making vast improvements here would definitely be valuable. But what if we want to do more detailed things like if/else blocks? Sometimes we call two senators and a committee and the scripts are subtlety  different for each one, but adding these subtleties gives people the confidence to call.

Does it make sense for us to expand this script templating language and implement it on each platform? Or make a roundtrip to the server asking for a script for each one? Or just include each script with a contact in the original issues request? Pros and cons for each, genuinely wondering what you think on the architecture here.

Perhaps additional thoughts from @oderby or @beausmith

nickoneill:
Closing this as it's based on the old codebase. Still interested in thoughts on making this something we can implement once across all platforms, I'll paste my notes into #310.

Issue #307, opened 2017-05-06
Fix zip/address lookup for Townhall data


nickoneill:
Previous solution used a large NPM module to do zip -> lat/lng lookups, which was a pain in development (es2040 translation takes more than a minute every time) and made the shipped app.js >5mb. I'm also not sure this solution supported distances from addresses.

We could potentially do a gmaps API geocoding lookup on address/zip while loading the townhall data in the background.

To start, revert this commit: 0de98b5e2d0a70f18f1d4d2de9a064455de38f2a

oderby:
Fixed in #314 right?

murribu:
yes

PR #314, opened 2017-05-15
Revert 0de98b5, remove zipcode lookup

murribu:
#307


oderby:
@murribu can you provide an overview of what you've changed in the code to avoid the problem which led to #307?

murribu:
Sure!
There is only one difference between this PR and #286:
In #286 , we would attempt to establish a user's lat/lng from their zipcode in LocalStorage. If the zipcode wasn't present, we would see if their lat/lng was in LocalStorage and use that.
Now in this PR, we will just skip the first step. The LocalStorage zipcode is ignored, and if the user has their lat/lng in LocalStorage, we'll use it.
If their lat/lng is not in LocalStorage, we will not show any THP events.

We can add more methods for establishing a user's location later (like calling the Google Maps API with the user's zipcode). I just wanted to limit this pull to _fixing_ the issue at hand.

static/js/components/townHall.js
```diff
@@ -0,0 +1,53 @@
+const html = require('choo/html');
+const t = require('../utils/translation');
+
+let renderLink = (e) => {
+  let linkExists = !!e.link;
+  if (linkExists){
```
oderby: Not sure this does anything to improve the legibility of the code over just `if (e.link) {` Since the if is only concerned with truthiness of a value, the double negation is not needed when used this way.

static/js/components/townHall.js
```diff
@@ -0,0 +1,53 @@
...
+    return html`<li><a href="${e.link}" target="_new">${t("townhall.moreInfo")}</a></li>`;
+  }
+};
+let renderNotes = (e) => {
+  let notesExist = !!e.Notes;
+  if (notesExist){
```
oderby: Same comment as above.

static/js/components/townHall.js
```diff
@@ -0,0 +1,53 @@
...
+    let ret = html`<li></li>`;
+    ret.innerHTML = e.Notes;
```
oderby: In general, this would be flagged as making our site vulnerable to XSS attacks - how much do we trust the content pulled from THP? Also, maybe it's not a huge deal, since we don't have any secure content/cookies on the client/browser.

That said, if there's a way we can do this safely, we should do it that way.

static/js/components/townHall_test.js
```diff
@@ -0,0 +1,663 @@
+const townHall = require('./townHall.js');
+const townHallUtils = require('../utils/townHallUtils.js');
+const chai = require('chai');
+const expect = chai.expect;
+const localEvents = [{
```
oderby: Can we pull this out to a separate file, townHall_test_data.json or such? Just to cleanup this file a bit.

static/js/main.js
```diff
@@ -142,10 +145,12 @@ app.model({
   reducers: {
     receiveActiveIssues: (state, data) => {
       const response = JSON.parse(data);
+      let divisions = townHallUtils.parseCivicData(response.divisions);
```
oderby: this can be `const`

static/js/components/townHall.js
```diff
@@ -0,0 +1,53 @@
...
+    return ret;
+  }
+};
+
+module.exports = (state) => {
+  if (state.localEvents && state.localEvents.length > 0){
```
oderby: nit: here and elsewhere, please insert a space between the `if` condidtion and open bracket. That is, style should be `if (foo === true) {`

static/js/main.js
```diff
@@ -253,10 +258,48 @@ app.model({
...
+        lat = geo.split(",")[0];
+        lng = geo.split(",")[1];
+      }
+      if (lat && lng){
+        events = townHallUtils.filterForLocalEvents(events, state.divisions, lat, lng);
+        // Only return (at most) the first three. Showing more than that could be overwhelming.
+        return {
+          localEvents: events.slice(0,3)
+        };
+      }else{
```
oderby: style nit: to be consistent with spacing in rest of document, use `} else {`

oderby: style nit: use `} else {` to be consistent with spacing in rest of doc.

static/js/main.js
```diff
@@ -253,10 +258,48 @@ app.model({
...
+    setCacheDate: (state, data) => ({ [data]: Date.now() }),
+    receiveTownHallData: (state, data) => {
+      let events = JSON.parse(data);
+      let lat = false;
+      let lng = false;
+      if (cachedGeo != ''){
+        // lat/long
+        let geo = cachedGeo.replace(new RegExp(/\]|\[|"/, 'g'),'');
+        lat = geo.split(",")[0];
+        lng = geo.split(",")[1];
```
oderby: Should probably protect here against the result of `geo.split` returning a length 1 array (will only happen if the cached geo location is borked, but could definitely happen). Or better yet, only use it if it's a length==2 array (anything else is malformed/wrong data type).

static/js/main.js
```diff
@@ -133,7 +134,9 @@ app.model({
     locationFetchType: cachedLocationFetchType,
     contactIndices: {},
     completedIssues: completedIssues,
-
+    localEvents: {},
```
oderby: `localEvents` appears to be an array, not an object. Should therefore be initialized to an empty array.

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,172 @@
...
+};
+let filterForLocalEvents = (events, divisions, lat, lng) => {
+  return events.map(mapDistanceAndDistrict(lat, lng))
+    .filter(filterEvents(divisions))
+    .sort(sortEvents);
+};
+module.exports = {
+  calculateDistance: calculateDistance,
+  parseCivicData: parseCivicData,
+  stateAbbrs: stateAbbrs,
```
oderby: I don't see anything using this...what's the reason to export it?

static/js/utils/townHallUtils.js
```diff
@@ -0,0 +1,172 @@
...
+    return R * c; // distance in miles
+  }else{
+    return null;
+  }
+};
+let deg2rad = (deg) => {
+  return deg * (Math.PI/180);
+};
+let getKeyByValue = (obj, val) => {
+  for( var prop in obj ) {
```
oderby: Just use `Object.keys` method, and then you can avoid the check on the following line entirely.

static/js/main.js
```diff
@@ -253,10 +258,48 @@ app.model({
...
+      let lat = false;
+      let lng = false;
+      if (cachedGeo != ''){
+        // lat/long
+        let geo = cachedGeo.replace(new RegExp(/\]|\[|"/, 'g'),'');
+        lat = geo.split(",")[0];
+        lng = geo.split(",")[1];
+      }
+      if (lat && lng){
+        events = townHallUtils.filterForLocalEvents(events, state.divisions, lat, lng);
```
oderby: When I run this code, I'm seeing the events data returned from THP endpoint is a JSON object, but `filterForLocalEvents` expects the `events` input to be an array.

oderby (changes requested):
The code doesn't run as is (see my comment at https://github.com/5calls/5calls/pull/314/files#diff-93471d4bf84d0569fadcebe5fba72773R273) . Also, some style cleanups and other organizational questions.

murribu:
I believe most of these issues have now been addressed. The exception is the XSS vulnerability. Perhaps we should have a backend endpoint to call which retrieves the THP data, sanitizes it, and returns it to the frontend?

static/js/utils/townHallUtils.js
murribu: CI yelled at me for using Object.keys. I guess IE doesn't like it.

static/js/components/townHall_test_data.json
```diff
@@ -0,0 +1,490 @@
...
+  "Member":"Bob Goodlatte",
+  "Notes":"<p>Held 1st Thursday every month. Rep. Goodlatte not in attendance; staff attends. More info here: <a href=\"http://goodlatte.house.gov/constituent-services/open-door-meetings.htm\" target=\"_blank\">Link</a></p>",
+  "Party":"Republican",
+  "RSVP":"",
+  "State":"Virginia",
+  "StateAb":"VA",
+  "Time":"11:30 AM",
+  "Zip":"22827",
+  "address":"20593 Blue and Gold Dr, Elkton, VA 22827",
+  "dateObj": (new Date()).getTime() + 100000, // make sure it's in the future
```
oderby: Oh, I didn't realize there was actual javascript code executed. This should actually be a js file then...

static/js/utils/townHallUtils.js
oderby: Yes, `Object.values` is not yet a fully supported operator. `Object.keys` should be fine to use though. Is there a reason you tried using `Object.values` below?

murribu: I got confused. I thought Object.keys wasn't supported and Object.values was supported. I figured Object.values would be a more straight-forward method of finding the stateAb than getKeyByValue. Since Object.keys is supported, I'm using that now.

static/js/main.js
```diff
@@ -253,10 +258,54 @@ app.model({
...
+        if (geo.split(",").length == 2) {
+          lat = geo.split(",")[0];
+          lng = geo.split(",")[1];
+        }
+      }
+      if (lat && lng){
+        eventsArr = townHallUtils.filterForLocalEvents(eventsArr, state.divisions, lat, lng);
+        // Only return (at most) the first three. Showing more than that could be overwhelming.
+        return {
+          localEvents: eventsArr.slice(0,3)
```
oderby, line 282: I think we should return the 3 closest events both in terms of location *and time* - right now, if there are more than 3 upcoming events at the same location, we just return a random 3 - I'd want to see the 3 upcoming ones.

static/locales/en.json
```diff
@@ -138,5 +138,10 @@
       "shareOnTwitter": "Share on Twitter",
       "shareOnFacebook": "Share on Facebook",
       "motto": "Spend 5 minutes. Make 5 calls. Make your voice heard."
+    },
+    "townhall": {
+      "findNearYou": "Find an upcoming event near you",
+      "moreInfo": "More info",
+      "thpCredit": "Events from the <a href='https://townhallproject.com' target='_new'>Town Hall Project</a>"
```
oderby, line 145: nit: I think it'd make sense to indicate that more information/events can be seen by going to the THP. Perhaps something like "These events and more information provided by the Town Hall Project"

oderby (changes requested):
Looking good. Just one important thing to address (returning consistently ordered events) and a minor copy suggestion to consider. Then lgtm.

static/js/main.js
murribu: Check out [townHallUtils.sortEvents](https://github.com/5calls/5calls/pull/314/files#diff-2b1285f6251ee2934b30977d2a72d8dbR149). It sorts events by distance - and if the distance is the same, it then sorts by date. So, I think this request is already addressed. Let me know if I'm misunderstanding the request.

static/locales/en.json
murribu: It would be fine with me to change this, but I'd like to also change the es.json accordingly. Shall I post it to the i18n slack channel?

static/js/main.js
oderby: Per conversation on slack, I think we should keep this sorting logic as is and revisit it after we get feedback/see uptake of feature.

static/locales/en.json
oderby: As discussed on slack, leaving copy as is for this initial version.

murribu:
So, is this ready to go?

oderby (approved):

nickoneill:
Awesome!

PR #326, opened 2017-06-22
WIP on tracker / custom outcomes

nickoneill:


static/js/components/outcomesButton.js
```diff
@@ -0,0 +1,23 @@
+const html = require('choo/html');
+const find = require('lodash/find');
+const t = require('../utils/translation');
+
+module.exports = (outcome, state, prev, send) => {
```
oderby, line 5: sigh, mixing this is not ideal, but assuming we are indeed moving to react, I'm inclined to let this slide for now. Because in react, you would pass the outcome in as a property, while it would be passed the state in the same manner it is now. So this approach is most aligned with what we'll do in react...

static/locales/es.json
```diff
@@ -111,10 +111,13 @@
...
     "outcomes":{
       "enterYourCallResult":"Meta el resultado de su llamada para obtener la siguiente llamada:",
-      "unavailable":"no disponible",
-      "voicemail":"dejé mensaje de voz",
-      "madeContact":"contacto hecho",
-      "skip":"omitir",
+      "outcome_unavailable":"no disponible",
+      "outcome_voicemail":"dejé mensaje de voz",
+      "outcome_contact":"contacto hecho",
+      "outcome_yes":"XXX",
```
oderby: TODO (just so we don't forget)

static/js/components/tracker.js
```diff
@@ -0,0 +1,52 @@
+const html = require('choo/html');
+const t = require('../utils/translation');
+const isEmptyObject = require('is-empty-object')
+
+module.exports = (state, prev, send) => {
+  console.log("trumpcare",state.ahcaCounts);
+
+  function normalizeName(name) {
```
oderby: please add a few unit tests for this.

static/js/components/tracker.js
```diff
@@ -0,0 +1,52 @@
...
+  if (isEmptyObject(state.ahcaCounts)) {
+    return html`Nah`;
+  } else {
+    var noCount = state.ahcaCounts.vm.length;
+    var yesCount = state.ahcaCounts.contacted.length;
+    var undecidedCount = state.ahcaCounts.unavailable.length;
+
+    return html`
+    <div class="tracker">
+      <p>Trumpcare is in the Senate and worse than ever. Id unum natum consul has, pri constituto efficiantur ut. Et harum officiis recteque vis, ut nec similique comprehensam, sed mazim denique definitiones ea. Help us keep this up-to-date by calling your representatives!</p>
```
oderby: TODO

static/js/components/tracker.js
```diff
@@ -0,0 +1,52 @@
...
+      <h3>Trumpcare Votezzz</h3>
+      <div class="tracker__votes">
+        <div class="tracker__votes__no" style="width:${noCount}%">No</div>
```
oderby: Do we want to translate these as well?

static/js/components/outcomes.js
```diff
@@ -9,28 +12,16 @@ module.exports = (state, prev, send) => {
...
 
   if (currentContact != null) {
     return html`<div class="call__outcomes">
       <h3 class="call__outcomes__header">${t("outcomes.enterYourCallResult")}</h3>
       <div class="call__outcomes__items">
-        <button onclick=${(e) => outcome(e, 'unavailable')}>${t("outcomes.unavailable")}</button>
-        <button onclick=${(e) => outcome(e, 'vm')}>${t("outcomes.voicemail")}</button>
-        <button onclick=${(e) => outcome(e, 'contacted')}>${t("outcomes.madeContact")}</button>
-        <button onclick=${(e) => outcome(e)}>${t("outcomes.skip")}</button>
+        ${outcomes.map((outcome) => outcomesButton(outcome, state, prev, send))}
```
oderby: instead of cloning above, why not just explicitly always add the skip outcome button here? (Fewer lines, avoid a clone)

static/js/components/tracker.js
```diff
@@ -0,0 +1,52 @@
...
+    // should turn something like "TX-TedCruz" into "Ted Cruz (TX)"
+    var stateRegex = /([A-Z]{2})-(.*)/;
+    var nameRegex = /([A-Z])([A-Z])([a-z])|([a-z])([A-Z])/g;
+    var match = stateRegex.exec(name);
+    var nameMatch = nameRegex.exec(match[2]);
+
+    return match[2].replace(/([A-Z])/g, ' $1')+" ("+match[1]+")";
+  }
+
+  if (isEmptyObject(state.ahcaCounts)) {
```
oderby: Do we really need another dependency to do this? Why not just `!state.ahcaCounts || Object.keys(state.ahcaCounts).length === 0`?

Or if you'd really like to use a third party for this, then just use `_.isEmpty` from [lodash](https://lodash.com/docs/4.17.4#isEmpty)

oderby (commented):
some cursory thoughts

static/js/components/outcomesButton.js
nickoneill: Ugh, yeah. I was not happy to end up with this, particularly after refactoring the script stuff to avoid it just the other day. I don't see another way to handle this in choo at least without tracking an index for which outcome we're working on in some global state.

static/js/components/tracker.js
nickoneill: I think that works fine, I just didn't know how to do it right.

nickoneill: Also how does one pick from the approximately 1000 tiny package options in npm

static/js/components/outcomes.js
nickoneill: You're right. Previously I was special casing it from `null` which would have been hard to pass from the server, now it just skips the key `skip` which can be sent.

static/js/components/tracker.js
```diff
@@ -1,52 +1,65 @@
...
+    <div class="tracker">
+      <h3>Trumpcare Votes</h3>
+      <p><strong>We'll be right back with updated vote tally counts.</strong></p>
+    </div>
+    `;
   } else {
-    var noCount = state.ahcaCounts.vm.length;
-    var yesCount = state.ahcaCounts.contacted.length;
-    var undecidedCount = state.ahcaCounts.unavailable.length;
+    var noVotes = [];
```
oderby: simplify: `var noVotes = state.ahcaCounts.no || [];`

static/js/components/tracker.js
```diff
@@ -1,52 +1,65 @@
...
+    if (state.ahcaCounts.unknown) {
+      unknownVotes = state.ahcaCounts.unknown;
+    }
 
     return html`
     <div class="tracker">
-      <p>Trumpcare is in the Senate and worse than ever. Id unum natum consul has, pri constituto efficiantur ut. Et harum officiis recteque vis, ut nec similique comprehensam, sed mazim denique definitiones ea. Help us keep this up-to-date by calling your representatives!</p>
-      <h3>Trumpcare Votezzz</h3>
+      <h2>Help Track Trumpcare</h2>
+      <p>Senate Republicans has kept us in the dark on Trumpcare, their new bill to replace the Affordable Care Act. This bill will affect one-sixth of our national economy, and the lives of every American.</p>
```
oderby: All this text should be translated...

static/locales/es.json
```diff
@@ -109,14 +109,24 @@
       "enterYourLocation":"<a class='location-link'>Introduzca su ubicacicón</a> para encontrar a sus representantes locales.",
       "setYourLocation":"<a class='location-link'> Establezca su ubicación </a> para encontrar a sus representantes"
     },
+    "tracker": {
```
oderby, line 112: TODO

package.json
```diff
@@ -53,6 +53,7 @@
     "i18next": "^7.1.1",
     "i18next-browser-languagedetector": "^1.0.1",
     "i18next-xhr-backend": "^1.3.0",
+    "is-empty-object": "^1.1.1",
```
oderby: Remove - not used anymore

static/js/components/outcomesButton.js
oderby: nod, i think this is fine as is, given my reasoning above.

static/js/components/tracker.js
```diff
@@ -0,0 +1,65 @@
+const html = require('choo/html');
+const t = require('../utils/translation');
+
+module.exports = (state) => {
+
+  function normalizeName(name) {
+    // should turn something like "TX-TedCruz" into "Ted Cruz (TX)"
```
oderby: Still want to see unit tests for this

static/js/components/tracker.js
```diff
@@ -0,0 +1,65 @@
+const html = require('choo/html');
```
oderby, line 1: This file really should have some tests.

static/js/components/tracker.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+    var unknownVotes = [];
+    if (state.ahcaCounts.unknown) {
+      unknownVotes = state.ahcaCounts.unknown;
+    }
+
+    return html`
+    <div class="tracker">
+      <h2>Help Track Trumpcare</h2>
+      <p>Senate Republicans has kept us in the dark on Trumpcare, their new bill to replace the Affordable Care Act. This bill will affect one-sixth of our national economy, and the lives of every American.</p>
```
oderby: Grammar: "Senate Republicans **have** kept us..."

oderby (commented):
tests? tests!

nickoneill:
Moved to #338

PR #331, opened 2017-06-29
moved the function receiveTownHallDataError into the reducers object

greggraf:
Because this function was not defined on the receiver object, it was not getting invoked on line 313 where the send() function is used.


nickoneill:
That explains some of js errors in sentry. Thanks!

nickoneill (approved):

PR #347, opened 2017-10-06
updated README to reflect that this is no longer the frontend repository

nlsandler:
Tell people that code now lives in react-dev, but clarify what this repo is still being used for.


nickoneill:
I think the failure here is due to the fork location, not anything material. Thanks for this update!

PR #355, opened 2017-11-04
Remove legacy issue migration

nickoneill:
This is the largest cause of js errors in sentry, not entirely sure why. At this point it makes more sense to remove than continue to migrate.


cdoremus:
We should also remove the code related to local storage migration in this PR too.

nickoneill:
@cdoremus that's what this is - is there another one I'm missing?

cdoremus (approved):
This looks good to go. I was mistaken about other migration code.

PR #359, opened 2017-11-06
added ReactMarkdown to Script component, with a simple test.

marshmallowrobot:


nickoneill (approved):
Very nice! I'm not super familiar with what can be tested but this is a great example of how we can use components (like <ReactMarkdown> here) to test itself, neat!

PR #361, opened 2017-11-17
Added support for non-contact action items

nickoneill:
Requested from one of our early teams, easy enough to test out.

Reworks outcome sending to be entirely the new `FlexibleOutcome` from the fetch contact type, no more enums :(

### DIFF
--- a/src/components/call/IssueLink.tsx
+++ b/src/components/call/IssueLink.tsx
@@ -2,16 +2,16 @@ import * as React from 'react';
 import { Issue } from '../../common/model';
 
 interface Props {
-  readonly issue: Issue;
+  readonly issue?: Issue;
 }
 
 const IssueLink: React.StatelessComponent<Props> = ({ issue }: Props) => {
-  let linkTitle = issue.link;
-  if (issue.linkTitle) {
-    linkTitle = issue.linkTitle;
-  }
-
-  if (issue.link) {
+  if (issue && issue.link) {
+    let linkTitle = issue.link;
+    if (issue.linkTitle) {
+      linkTitle = issue.linkTitle;
+    }
+  
     return (
       <h4 className="call__script__link"><a target="link" href={issue.link}>{linkTitle}</a></h4>
     );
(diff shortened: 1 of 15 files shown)


nickoneill:
Example is `team/nick/32`
@nickoneill Get a 'NO CALLS TO MAKE' message when I browse to this issue.

cdoremus (changes requested):
@nickoneill for the record, this is a summary of my recommendations I discussed with you on Slack last night:

I've done some experimenting with the PR and it appears that you can use `OutcomeData` instead of `FlexibleOutcomeData`, by having the `outcome` field of this type: `UserContactEventType | string` This allows you to use code completion if you want the old outcomes or add a custom outcome as a string. In either case, a string is passed to the back end as the outcome value.

You don't need that switch statement in `Outcomes.dispatchOutcome()` that you removed in the PR. You can also remove `FlexibleOutcomeData` and replace all references to it with `OutcomeData`.

To keep things cleaner, I would create a new type:
```
type UserOutcomeResult =  UserContactEventType | string;
```

The call signature of dispatchOutcome becomes
```
  dispatchOutcome(e: React.MouseEvent<HTMLButtonElement>,
                  outcome: UserOutcomeResult = UserContactEventType.SKIP) 
```

You'll also need to change `UserContactEvent.result` to be of type `UserOutcomeResult`.

nickoneill:
Thanks! This'll make things much simpler.

src/components/call/IssueLink.tsx
```diff
@@ -2,16 +2,16 @@ import * as React from 'react';
...
-
-  if (issue.link) {
+  if (issue && issue.link) {
+    let linkTitle = issue.link;
+    if (issue.linkTitle) {
+      linkTitle = issue.linkTitle;
+    }
+  
     return (
       <h4 className="call__script__link"><a target="link" href={issue.link}>{linkTitle}</a></h4>
```
cdoremus, line 16: If the `issue` prop is optional (as defined on line 5),  `issue.link` will throw a `ReferenceError` here when it is `undefined`.

nickoneill: Not in this case, it will fail at `issue` first and won't continue to evaluate the cases. Otherwise this would fail when you reload a page from scratch in dev (there's initially no currentIssue and the page has to search the downloaded issues for something that matches the path)

cdoremus (approved):
Good to go. Nice job!
Onward and upward!!

PR #362, opened 2017-11-20
Added a query param to force the location.

nickoneill:
This is coming in handy with automating something I'm working on, but could also be used to send people to the right districts if they're split and don't know how to fix.

### DIFF
--- a/src/redux/remoteData/asyncActionCreator.ts
+++ b/src/redux/remoteData/asyncActionCreator.ts
@@ -9,6 +9,7 @@ import { issuesActionCreator, groupIssuesActionCreator, callCountActionCreator }
 import { clearContactIndexes } from '../callState/';
 import { ApplicationState } from '../root';
 import { LocationUiState } from '../../common/model';
+
 /**
  * Timer for calling fetchLocationByIP() if
  * fetchBrowserGeolocation() fails or times out.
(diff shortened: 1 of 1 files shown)


src/redux/remoteData/asyncActionCreator.ts
```diff
@@ -211,6 +211,19 @@ export const startup = () => {
...
     dispatch(clearContactIndexes());
 
+    // if a location is passed as a query, override or set the location address manually
+    let addressQuery = 'forceAddress';
+    let query = window.location.search.substring(1);
+    let vars = query.split('&');
+    for (let i = 0; i < vars.length; i++) {
+        let pair = vars[i].split('=');
+        if (decodeURIComponent(pair[0]) === addressQuery) {
+          state.locationState.address = pair[1];
```
cdoremus: State always needs to be updated by using a Redux reducer via an Action created by an ActionCreator.

cdoremus (changes requested):
As stated inline, locationState needs to be updated to dispatch an Action to a Redux reducer via an action creator.

src/redux/remoteData/asyncActionCreator.ts
nickoneill: It didn't look like `setState` is available here. How would I go about getting that in this context?

cdoremus: Use the dispatch function to call the appropriate action creator.

nickoneill: Ohhhh right

nickoneill:
OK! Thanks for pointing that out, makes total sense.

cdoremus (approved):
Thanks for making the changes.

PR #364, opened 2017-11-23
A proper not found page if people use a url that isn't routed

nickoneill:
This is better than serving the home page for every url, right?

Note: I combined the idea of a Container and Page into a single file for this since it's unlikely it grow very large.


cdoremus:
@nickoneill  this is a good idea, but I think that `NotFound` should be called `NotFoundPage` to be consistent with the other pages (knowing that `NotFound` encapsulates a Page, Container and Component).

nickoneill:
OK, that makes sense.

nickoneill:
ok, all set!

cdoremus (approved):
Good to go!

PR #365, opened 2017-11-25
Updated some packages to react 16

nickoneill:
Couple things complaining when doing a fresh install, and snapshot tests failing related to an enzyme serializer. Sounds like this is fixable: https://github.com/rogeliog/jest-serializer-enzyme/issues/15

(I was testing this because of some static work I'm doing, I heard that react 16 "upgrades" from static html to hydrated react html more cleanly but I honestly don't see a difference in the render. I am serving html with the proper issue html, but still get a flash of "no calls to make" before the issue title appears again.)

What else needs to be done here for upgrading to be "done?". Other than the obvious like getting the tests running happily again.


nickoneill:
Updated enzyme for react 16 and had to do a little dance to get it working again. Looks good now though!

cdoremus:
It sounds like version 16.2 would be a good upgrade (see [https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html)).

nickoneill:
Happy to make that upgrade as well but we aren't using either the array of elements (which seems clunky and unusual) or the fragments anywhere at the moment.

nickoneill:
Updated to 16.2.0 for fragment support, rebased on master

Issue #367, opened 2017-12-06
Group info caching


nickoneill:
I think it would be helpful to groups** to show their header (group name, logo, etc) on the top of their call pages as well as the group home page.

Problem is, we only know the group id via the url param and not the group name, so we go fetch that from the server. The roundtrip makes the experience pretty meh.

Let's investigate a way to cache the group info, possibly at `RemoteDataState` for a certain amount of time. This should not just be the current group, but any group we navigate to, so a map from groupid -> group object.

* This could be cached for a certain amount of time, or even just updated in the background every time you visit the group home page.
* We shouldn't refetch the group info on each group call page, but we should fetch it a single time if this is the first time you've visited the group at all and don't have it cached.
* This would also enable us to show the group name in the sidebar rather than "your group" like we do now. That's a much better experience.

** The 5 Calls webapp has the functionality for individuals to form groups (or teams) that allows them to create their own issues (in a separate admin interface) and keep track of the group's call statistics.

cdoremus:
For reference, here's the current `RemoteDataState` structure:
```
export interface RemoteDataState {
 issues: Issue[];
 inactiveIssues: Issue[];
 currentGroup: string;
 groupIssues: Issue[];
 callTotal: number;
 donations: Donations;
 errorMessage: string;
}
```
And here is the current `Group` interface:
```
export interface Group {
  id: string;
  name: string;
  subtitle: string;
  description: string;
  totalCalls: number;
  photoURL: string;
}
```

cdoremus:
Here's my current thinking on this issue:
1. Move `pageGroup` from `GroupPage` local state to `CallState` app state (also add a timestamp field to the `Group` interface).
-- `CallState` is persisted in local storage (`RemoteDataState` is not). 
-- `pageGroup` becomes a `GroupPage` prop.
2. At startup when `startup()` is called, the call state's async action creator is invoked to call `apiServices.getGroup()` (an API call) and set `CallState.pageGroup` by dispatching an action to call state's reducer.
-- the API call would set the `Group` timestamp to the current timestamp.
3. `GroupPageContainer.mapStateToProps()` is used to set the `pageGroup` property using `CallState.pageGroup`.
-- Stale data is checked using `Group.timestamp` (and current timestamp) in `GroupPageContainer.mapStateToProps()`. If the data is stale, the API call is done to refresh `CallState.pageGroup`.
[...]

PR #369, opened 2017-12-12
Implement group/team info caching per issue #367

cdoremus:
This PR implements an application cache key in the Redux store to hold group/team data that is persisted using `redux-persist`. It is an implementation of issue #367.

### DIFF
--- a/src/components/call/FetchCall.tsx
+++ b/src/components/call/FetchCall.tsx
@@ -4,7 +4,7 @@ import i18n from '../../services/i18n';
 import { TranslationFunction } from 'i18next';
 import * as ReactMarkdown from 'react-markdown';
 
-import { Issue, VoterContact, Group } from '../../common/model';
+import { Issue, VoterContact, CacheableGroup } from '../../common/model';
 import { CallHeaderTranslatable } from './index';
 import { CallState, OutcomeData } from '../../redux/callState';
 import { LocationState } from '../../redux/location/reducer';
@@ -14,7 +14,7 @@ import { queueUntilRehydration } from '../../redux/rehydrationUtil';
 // This defines the props that we must pass into this component.
 export interface Props {
   readonly issue: Issue;
-  readonly currentGroup?: Group;
+  readonly currentGroup?: CacheableGroup;
   readonly callState: CallState;
   readonly locationState: LocationState;
   readonly t: TranslationFunction;
--- a/src/components/call/FetchCall.tsx
+++ b/src/components/call/FetchCall.tsx
@@ -4,7 +4,7 @@ import i18n from '../../services/i18n';
 import { TranslationFunction } from 'i18next';
 import * as ReactMarkdown from 'react-markdown';
 
-import { Issue, VoterContact } from '../../common/model';
+import { Issue, VoterContact, Group } from '../../common/model';
 import { CallHeaderTranslatable } from './index';
 import { CallState, OutcomeData } from '../../redux/callState';
 import { LocationState } from '../../redux/location/reducer';
@@ -14,7 +14,7 @@ import { queueUntilRehydration } from '../../redux/rehydrationUtil';
 // This defines the props that we must pass into this component.
 export interface Props {
   readonly issue: Issue;
-  readonly currentGroupId?: string;
+  readonly currentGroup?: Group;
   readonly callState: CallState;
   readonly locationState: LocationState;
   readonly t: TranslationFunction;
@@ -119,7 +119,7 @@ export default class FetchCall extends React.Component<Props, State> {
     if (this.supportEnabled()) {
       return (
       <div className="call__outcomes__items call__outcomes__support">
-        {buttons.map((button, index) => 
+        {buttons.map((button, index) =>
           <button key={index} onClick={(e) => this.setSupport(e, button.key)} className={this.buttonClass(button.key)}>
             {button.title}<br/>{button.emoji}
           </button>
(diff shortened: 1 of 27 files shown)


cdoremus:
@nickoneill what timeout value would you suggest for the cache?

nickoneill:
not sure yet, let's start at 24h

nickoneill:
How's this going @cdoremus? Anything I can do to help?

cdoremus:
@nickoneill Sorry, but I've been distracted by job search and holiday responsibilities. Still, I have done some work that I should be able to push this evening or tomorrow.

nickoneill:
No worries! Just checking in. Let me know if I can help or review.

cdoremus:
The commit I pushed today has failing tests because of some `console.log()` calls I added for debugging purposes. These will be removed when the code is ready for review. If you run the tests using `yarn test:nolint`, all tests pass.

cdoremus:
Making progress... 

I still need to change `currentGroupId: string` prop to `currentGroup: Group` to allow display of group name in the `SidebarHeader` and manually test some corner cases.

The unit tests now pass.

nickoneill:
🎇

nickoneill:
Awesome! I am just starting to go through this and my first pass worked fine, but deleting the appCache object gets me stuck into this state where my group isn't cached and it won't refetch.

I'm trying to figure out where the cache timestamp is stored but I can't figure it out - how have you been resetting this to try it fresh?

cdoremus:
@nickoneill How are you deleting the `appCache` object? I want to be able to reproduce the bug you found.

All group data gets stored in the `reduxPersist:appCache` key in local storage. When I delete that key, the app continues to function OK. The `appCache.groups` object contains an array of `CacheableGroup` objects. Each `CacheableGroup` contains properties `group: Group` and `timestamp: number` (milliseconds). 

When a group is requested, the `appCache.groups` array is searched by group id and if a group is found  the timestamp of the `CacheableGroup` that it is in iis checked against the current timestamp to see if the cached object has timed out. (see `redux/cache/asyncActionCreator.cacheGroup()`). If the group has timed out, an API call is made to update the group object of the `CacheableGroup` and the timestamp is set to the current timestamp. 

If a group is not found a new Group is created from an API call and added to a new `CacheableGroup` with a current timestamp. This new `CacheableGroup` is then added to the `appCache.groups` Redux state object.

I looked at the code again and have started refactoring the `GroupPageContainer` and `GroupPage` to clarify and simplify things. I also need to make sure the cache timeout is checked in `GroupPage`. I hope to check this stuff in as soon as time allows.

nickoneill:
I just deleted the `reduxPersist:appCache` key in localstorage through chrome. It populates again when I load another group page but it's empty (`{"groups":[]}`), it never seems to pick up that particular group information.

cdoremus:
@nickoneill I just discovered I need to change all the `currentGroup` references throughout from type `Group` to `CachableGroup`. Expect another commit for this, but please review what I committed today. TIA

cdoremus:
I've done the refactoring and everything should be ready to go (I think :smiley_cat: ).

I have to say that TypeScript really makes these kind of refactorings easy!

src/components/call/FetchCall.tsx
```diff
@@ -250,15 +250,15 @@ export default class FetchCall extends React.Component<Props, State> {
...
-      return <h3 className="call__outcomes__header">Getting your next contact...</h3>;      
+      return <h3 className="call__outcomes__header">Getting your next contact...</h3>;
     } else {
       return (
         <blockquote>
           <h2 className="call__outcomes__header">All done for today!</h2>
           {/*tslint:disable-next-line:max-line-length*/}
           <p>Looks like we're all out of calls to make for today, or we're outside of normal calling hours (9am-9pm in the local time zone). Come back tomorrow for more calls!</p>
         </blockquote>
-      );
```
nickoneill, line 126: Sorry about these, I don't know why VS code doesn't trim them like any sane software

src/components/groups/GroupPage.tsx
```diff
@@ -26,63 +24,84 @@ interface Props extends RouteProps {
...
-    if (this.state.pageGroup) {
-      if (newProps.match.params.groupid !== this.state.pageGroup.id) {
-        this.setState({ loadingState: GroupLoadingState.LOADING });
-        this.getGroupDetails(newProps.match.params.groupid);
-      }      
+  setGroup(props: Props) {
+    let cgroup = this.state.pageGroupState;
+    // Call the API when page first loads and group
+    // is not in the cache or if the cached value has timed out
+    if (!cgroup || hasGroupCacheTimeoutExceeded({timestamp: cgroup.timestamp})) {
```
nickoneill, line 62: so the page is responsible for expiring a group? Feels like this might be duplicated on multiple pages in the future (whenever we use a group) but probably OK for the moment.

Not sure where I can imagine this going. Since the getter for groups is really just the state object and not some function that you can insert timeout checking.

nickoneill (changes requested):
Overall great stuff! One specific comment inline without any particular changes requested, but curious on your opinion.

One thing I noticed: if you have no cached groups and you hit a call detail page (`team/nick/2`), the group isn't fetched and cached. The sidebar still lists the groups issues but the links are all broken since there's no actual group id to construct a proper group url.

Ideas here? This might touch on that other code comment I had as well now that requesting the group cache is on more than one page.

nickoneill:
Gonna resolve the conflict here and pull this in as a start so I can base some future work on this.

nickoneill:
Why I had to merge the same conflict twice is beyond me.

PR #374, opened 2017-12-20
ACA Outcomes and some maintenance updates

nickoneill:


src/components/call/ACAOutcomes.tsx
```diff
@@ -0,0 +1,235 @@
...
+  render() {
+    return (
+      <div className="call__outcomes">
+        <h3 className="call__outcomes__header">
+          How did the call go?
+        </h3>
+        <div className="call__outcomes__items">
+          {outcomeButtons.map((button, index) =>
+            <button
+              key={index}
```
cdoremus: Use `button.key` as the key attribute (see article [Index as a key is an anti-pattern](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318)). When that is done, remove the `index` argument from the `outcomeButtons.map()` call.

src/common/model.ts
```diff
@@ -167,3 +167,9 @@ export interface Donations {
     total: number; // goal (api===total)
     kind: string; // denomincation (dollars)
 }
+
+export interface Button {
+  title: string;
+  emoji: string;
+  key: string;
```
cdoremus: I suggest to type key as `key: UserOutcomeResult` (`UserOutcomeResult` is defined in `redux/userStats/reducer.ts`). This allows for code reuse and consistent outcomes (for instance, `contacted` will become `contact` and `nothome` will become `unavailable`). The JavaScript value of a string-based TypeScript enum will be the string.

When doing this refactoring, you should move `UserOutcomeResult` and `UserContactEventType` to `model.ts`. I also think that `UserOutcomeResult` should be renamed `OutcomeResult` and `UserContactEventType` renamed to `StandardOutcomeEvent` (or something else that reflects their more generic use).

src/common/model.ts
```diff
@@ -167,3 +167,9 @@ export interface Donations {
     total: number; // goal (api===total)
     kind: string; // denomincation (dollars)
 }
+
+export interface Button {
```
cdoremus: You might want to call this interface `OutcomeButton` that is more reflective of its intention.

nickoneill: Yeah, definitely more clear

src/common/model.ts
nickoneill: This is a good set of improvements but there's more thinking that needs to go into how these buttons are created and sent to the client because of upcoming requirements. Let's deal with this as part of that refactor.

src/components/call/ACAOutcomes.tsx
nickoneill: 👍

nickoneill:
Pulling this in so I can get the tracker removed from the front-page today. Will note the other outcome model improvements for refactors as mentioned.

PR #375, opened 2017-12-21
Added a new section to faq, made a few style tweaks

### LINKED ISSUE #373: add item to FAQ to explain why it's still good to call your reps if they are doing what you want
people write in a lot to ask this, and i just had a friend ask too, and suggest we put it in the faq : https://5calls.org/faq
here is some proposed text: 

**If my representation is doing the right thing, should I still call?**
Yes! Your legislators need to know you support them, and they often use the tally of calls from their constituents to support their position when speaking to their colleagues in Congress. It's particularly helpful in some cases for them to have personal stories from constituents to add to their discussion, so if something affects you or your family and friends directly, share your story when you call!

nickoneill:
Fixes #373


piebob (approved):
LGTM :D 👍

PR #377, opened 2017-12-31
Added small screen layout for donation section

cdoremus:
When displayed on a cell phone the donation section on the home page takes up a large amount of vertical screen real estate. This PR compresses the donation section on a small screen. See the screen-shot below for a comparison of the current 5calls.org donation section on a small screen and the one created by this PR (edits to the `__layout.scss` file).

donationsectioncomparisonsonasmallscreen_currentwebsite-vs-thispr

### DIFF
--- a/src/components/shared/scss/_layout.scss
+++ b/src/components/shared/scss/_layout.scss
@@ -40,6 +40,10 @@
   min-height: 12px;
   border-bottom: solid 8px $red;
 
+  @include bp(small) {
+    border-bottom: solid 2px $red;
+  }
+
   &__logo {
     position: relative;
     padding: 10px 30px 0px 30px;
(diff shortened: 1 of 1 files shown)


nickoneill:
Agree with the premise. A couple style notes:

* Let's keep the header "5 calls is powered by donations from..." centered
* Keep the rounded corners on the buttons
* Swap the side of the button: button on the right, "supports" text on the left
* Even out the spacing so the red space below the buttons is the same height as the space between button sections

cdoremus:
@nickoneill I just pushed a few tweaks that I think addresses your concerns (see below).

screenshot from 2017-12-31 22-46-49

nickoneill (approved):
👍

PR #408, opened 2018-07-20
Making the done page a bit nicer

nickoneill:
* Moves the donation prompt from the header to the done page
* Adds a new share prompt which has an image preview
* Adds an experimental phone reminder feature

Also:
I think we weren't refreshing the user profile correctly, so that should be better here. Plus fixed a small issue with geolocation and google civic.


nickoneill:
I deployed this because it includes some important login refresh and location fixes, but I'm hesitating merging it because the tests are failing on some new phone number validation code.

The error in particular is `Cannot find module 'libphonenumber-js/metadata.min' from 'PhoneInputReactResponsiveUIDefaultMetadata.js'` which is referenced here: https://github.com/catamphetamine/libphonenumber-js/issues/199

The "fix" is including `.json` files in the webpack config, which we already have so I can't figure out why this is still failing on tests. Presumably the `metadata.min.json` file has some phone validation data inside it that the module uses and for some reason isn't being included when testing.

@nabrahamson do you have any webpack or jest experience that could help solve this?

nickoneill:
I feel bad about this. It is getting merged this way because I cannot spend another minute on obscure webpack errors.

PR #409, opened 2018-07-20
Add Netlify config

nickoneill:


nickoneill:
Currently this adds the netlify config so it knows what to serve. We can merge this in whenever and start a new branch that netlify can deploy automatically.

PR #411, opened 2018-08-01
Remove uneeded calls to /issues?address

nabrahamson:


nickoneill:
So this fully fixes the multiple request issue? What was the interaction between `fetchGroupIssuesIfNeeded` and `fetchIssuesIfNeeded` that caused the issue?

nabrahamson:
From that I understand, componentWillRecieveProps is called whenever props for the component is changed. As the props were being injected into the component in rapid succession, it would trigger the call to issues? Location before it had a chance to finish. 

I ran all available tests after making this change as well as testing in my dev machine clearing the cache after each attempt to compare current functionality against functionality before my changes.

nabrahamson:
One more note, componentWillRecieveProps is depreciated and marked unsafe and should be removed in future refactoring.

nickoneill:
I'll check out how this works a little later. This *might* break functionality with groups who want to use the standard 5 Calls issues rather than write custom ones, but I don't know if any of those groups are actively using their pages (custom calls is more popular).

At any rate, we could move this to the spot where the custom group issues are loaded and switch there if it's super important.

nabrahamson:
I could also put that piece back and add another conditional to only call that block of code if this.props.issues has changed.

nickoneill:
This still looks like it's happening on my end, try loading this url in dev: 
`http://localhost:3000/team/nick/152`
screen shot 2018-08-02 at 1 24 21 pm

nabrahamson:
I've removed the similar pattern from the CallPage

nabrahamson:
The fail on the cirleci check was an error trying to download a package. I believe that if we run that test again, it should pass.

nickoneill:
Let's wait on this... I want to adjust how we get contacts and issues more broadly and we'll have to refactor this part anyhow.

Issue #395, opened 2018-04-05
Add state legislator titles to script formatter


nickoneill:
Our script formatter adds proper senator or representative titles to national reps when the writers add [SEN/REP NAME] to scripts, but doesn't properly give titles for state legislators. Let's add this to the script formatter.

Since we can't (easily) tell if the state rep is an assemblyperson, or state senator or whatever, let's use the term "Legislator [Lastname]"

itbeauregard:
Right now, it looks like it actually just prefixes Senator to everyone.

nickoneill:
@itbeauregard We target more than just Senate and House reps (and local legislators) here, so if possible it would be nice to ONLY replace the string when we know we have an accurate title for that contact. If it's a contact type that we don't know about, leave the [SEN/REP NAME] without replacement.

itbeauregard:
Update: the `contact` data object has an `area` property that can determine whether a legislator is state rep or state senator. The identifications seem to work as follows:
"House" means National Rep
"Senate" means National Senator
"StateLower" means State Rep
"StateUpper" means State Senator

@nickoneill Do you want me to give the local reps and senators their title or stick with Legislator? I can ensure that everyone else doesn't get a prefix.

nickoneill:
"Legislator" for both State Upper and State Lower, please. The legislature has different names based on what state its in and this is a good catch-all term.

itbeauregard:
Got it. Will change.

nickoneill:
Here's the current list of "areas" we define for our calls. Again, there are some that fall outside this bucket which should have no title, but this is the current list of areas we frequently use:

```
	areaHouse            = "House"
	areaSenate           = "Senate"
	areaGovernor         = "Governor"
	areaAttorneyGeneral  = "AttorneyGeneral"
	areaSecretaryOfState = "SecretaryOfState"
	areaStateUpper       = "StateUpper"
	areaStateLower       = "StateLower"
```

If possible, we could add "Governor X", "AG X" and "Secretary of State X" cases as well.

itbeauregard:
In PR [here](https://github.com/5calls/5calls/pull/412)

PR #412, opened 2018-08-03
Issue 395: Ensure state legislators are addressed correctly

itbeauregard:
Addresses [Issue 395](https://github.com/5calls/5calls/issues/395) which is concerned with state legislators having the correct prefix. Right now, all state reps and senators get the "Senator" prefix. 

The `currentContact` object has an `area` property that can determine whether a legislator is state rep or state senator. The identifications seem to work as follows:
"House" means National Rep
"Senate" means National Senator
"StateLower" and "StateUpper" means Local Legislator

@nickoneill has said that we target other non-legislators, so we are leaving those folks without a prefix. 

Will add test for `getContactNameWithTitle`.


nickoneill (approved):
This looks fantastic @itbeauregard! Thanks for writing the tests as well!

nickoneill:
I'll merge this and deploy with a few other changes later tonight.

PR #413, opened 2018-08-03
Improve accessibility of CallCount progress bar

aloverso:
I noticed that the text contrast (white on light blue) in the CallCount progress bar did not adhere to the [WebAIM guidelines](https://webaim.org/resources/contrastchecker/)

This tweak to CallCount styling updates the style of the progress bar by moving the label text (goal and current count) to be labels outside of the progress bar itself.

Additionally, this commit includes a refactor of the logic in the CallCount that computes the current goal to display.  It was previously hardcoded to a list of possible values.  Now, it computes the goal dynamically by determining the next power of ten (10,100,1000, etc) to shoot for, or the halfway point (50,500,5000, etc).

This way, as the site usage grows, the logic will grow with it and no updates will be needed when the progress bar continues to advance.

This seemed like a small refactor / accessibility fix that did not require opening an issue through the product guidelines.

### DIFF
--- a/src/components/shared/CallCount.tsx
+++ b/src/components/shared/CallCount.tsx
@@ -11,54 +11,32 @@ export interface Props {
 }
 
 export const CallCount: React.StatelessComponent<Props> = (props: Props) => {
-  let callGoal = 100;
-  if (props.totalCount >= 100) {
-    callGoal = 500;
-  }
-  if (props.totalCount >= 500) {
-    callGoal = 1000;
-  }
-  if (props.totalCount >= 1000) {
-    callGoal = 5000;
-  }
-  if (props.totalCount >= 5000) {
-    callGoal = 10000;
-  }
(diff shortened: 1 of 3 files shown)


nickoneill:
Thanks @aloverso, this is a great change!

Looks like the circle failure is something intermittent on their side. I've restarted the tests, sometimes you have to rerun them a few times to get it to complete 🤔

nickoneill:
For reference, here's the updated look:
screen shot 2018-08-07 at 2 08 16 pm

I'm not sure the highly rounded corners and white border integrates with the existing design elements. What do you think about scaling back the design changes a bit, keeping your improvements to the webaim guide, something similar in look to this mock?
call total mock

nickoneill:
Hi @aloverso - any thoughts on the suggested change?

nickoneill:
We'll take a look at this again with the new homepage design in #460

PR #416, opened 2018-08-13
Opening PR for WIP

nabrahamson:
Implement context api for groups

THIS CURRENTLY DOES NOT RUN
There is a dependency conflict with 5-calls/react-components


package.json
```diff
@@ -5,7 +5,9 @@
   "private": true,
   "dependencies": {
     "@5calls/react-components": "^0.0.10",
-    "axios": "^0.16.2",
+    "@types/redux": "^3.6.0",
```
nickoneill: Anything under @types should be dev dependencies only

src/common/model.ts
```diff
@@ -87,6 +89,25 @@ export interface Group {
...
+    this.name = '';
+    this.subtitle = '';
+    this.description = '';
+    this.totalCalls = 0;
+    this.photoURL = '';
+    this.customCalls = false;
+    this.subscribed = false;
+  }
+
+  static from(groupState: GroupState) {
```
nickoneill, line 104: Does this make sense here? Or on `GroupState`?

src/components/layout/Header.tsx
```diff
@@ -41,9 +41,16 @@ class HeaderImpl extends React.Component<Props, State> {
   }
 
   render() {
-    let profile: UserProfile | undefined;
-    if (this.props.currentUser !== undefined) {
+    let profile: UserProfile;
+    if (this.props.currentUser !== undefined && this.props.currentUser.profile !== undefined) {
       profile = this.props.currentUser.profile;
+    } else {
```
nickoneill: Having an empty profile isn't the same as the profile being undefined. Do we check this up the stack or something?

src/contexts/RemoteStateContext.ts
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import {
+  Group,
+  Issue,
+} from '../common/model';
+
+export interface ProviderState {
```
nickoneill: If this is the `RemoteStateContext` doesn't that mean the interface should be `RemoteState` rather than `ProviderState`?

src/redux/group/asyncActionCreator.ts
```diff
@@ -11,7 +11,8 @@ export const updateGroup = (group: Group) => {
     ) => {
     const state = getState();
     if (state.groupState.groupLoadingStatus === GroupLoadingActionStatus.LOADING) {
-      dispatch(cacheGroup(group.groupID));
+      // tslint:disable-next-line:no-any
+      dispatch<any>(cacheGroup(group.groupID));
```
nickoneill: Does dispatch complain without a type here? I don't think I have to do the same in 5calls-action.

nickoneill (commented):
Some general questions

src/common/model.ts
nabrahamson: I think this makes more sense here as it's creating a Group object from a GroupState interface. I believe putting it on GroupState would require us to convert GroupState to a class.

src/components/layout/Header.tsx
nabrahamson: I think this is related to the current login issues we've talked about over Slack. Looking at the react-components repo, the userProfile param is defined as optional, so I shouldn't need to pass in an empty profile.

nickoneill: Agreed. Do you recall why you needed the empty profile here when this change was made?

nabrahamson: Figured out the login error but I need to figure out why it occurred. For some reason when I pull down the react-components repo, the typings defininition for the CustomLogin component isn't defining the userProfile prop as optional. If I go into the node modules folder and edit the typings definition for the component, I can get rid of the empty profile and things work as expected.

nabrahamson: Building the current version of react-components on my machine gives the correct typings definition for this component. This block of code will be reverted in my next push.

nabrahamson:
To reply to the comment about Calls.tsx. I changed it to slug because the dictionary contactIndexes was keyed on issue slugs. I can change the logic so that it will look for the id first, then the slug in case the id returns undefined.

src/common/constants.ts
```diff
@@ -50,3 +52,5 @@ export const contact = {
   facebook: 'https://www.facebook.com/make5calls',
   apps: 'https://crgj.app.link/7R2bEB0R4F',
 };
+
+export const PAGETITLE = '5 Calls: Make your voice heard';
```
nickoneill: Did you add this with the intent of using it somewhere? Search doesn't bring it up elsewhere on this page.

src/common/model.ts
nickoneill: OK, fair enough

src/components/AppProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+  groupStateContext,
+  userStateContext,
+  userStatsContext,
+  appCacheContext,
+} from '../contexts';
+
+interface Props {
+  // temporary until completely converted to Provider pattern
+  // tslint:disable-next-line:no-any
+  store: Store<any>;
```
nickoneill: I believe this can be `Store<ApplicationState>` now that everything is in

src/components/NotFoundPage.tsx
```diff
@@ -48,9 +54,7 @@ const NotFoundPage: React.StatelessComponent<AllProps> = (props: AllProps) => {
           <p>Looks like you visited a page that doesn't exist. Pick one of the issues on the sidebar or <Link to="/">go back to the homepage</Link>.</p>
         </main>
       </div>
-      <Footer
-        t={i18n.t}
-      />
+      <Footer/>
```
nickoneill, line 57: How does this work now that t isn't sent to footer and to navigation below that? Navigation is using `props.t` to translate stuff, and it appears correctly. I'm not 100% on where that comes from since we removed it from the props passed to the footer.

src/components/call/ContactDetails.tsx
```diff
@@ -8,7 +8,8 @@ import { makePhoneLink } from '../shared/jsxUtils';
 interface Props {
   readonly currentIssue: Issue;
   readonly contactIndex: number;
-  readonly t: TranslationFunction;
+  // tslint:disable-next-line:no-any
+  readonly t: TranslationFunction|any;
```
nickoneill: Let's remove this `any`

src/components/call/ContactOffices.tsx
```diff
@@ -7,11 +7,12 @@ import { makePhoneLink, cityFormat } from '../shared/jsxUtils';
 interface Props {
   readonly currentIssue: Issue;
   readonly contactIndex: number;
-  readonly t: TranslationFunction;
+  // tslint:disable-next-line:no-any
+  readonly t: TranslationFunction|any;
```
nickoneill: Remove this `any`

src/components/call/NoContactSplitDistrict.tsx
```diff
@@ -2,17 +2,19 @@ import * as React from 'react';
...
 import { translate } from 'react-i18next';
+import { clearAddress } from '../../redux/location';
+import { store } from '../../redux/store';
 
 interface Props {
   readonly splitDistrict: boolean;
-  readonly clearLocation: () => void;
-  readonly t: TranslationFunction;
+  // tslint:disable-next-line:no-any
+  readonly t: TranslationFunction|any;
```
nickoneill: Remove `any`

src/components/call/Outcomes.tsx
```diff
@@ -2,15 +2,16 @@ import * as React from 'react';
...
 
-interface Props {
-  readonly currentIssue: Issue;
-  readonly currentContactId: string;
+interface Props { readonly currentIssue: Issue; readonly currentContactId: string;
   readonly numberContactsLeft: number;
-  readonly t: TranslationFunction;
-  readonly onSubmitOutcome: (data: OutcomeData) => void;
+  // tslint:disable-next-line:no-any
+  readonly t: TranslationFunction|any;
```
nickoneill: `any`

src/components/call/Script.tsx
```diff
@@ -8,7 +8,8 @@ import { LocationState } from '../../redux/location/reducer';
 interface Props {
   readonly issue: Issue;
   readonly contactIndex: number;
-  readonly t: TranslationFunction;
+  // tslint:disable-next-line:no-any
+  readonly t: TranslationFunction|any;
```
nickoneill: `any`

src/components/done/DonePage.tsx
```diff
@@ -1,25 +1,33 @@
...
+  getIssuesIfNeeded,
+  RemoteDataState,
+} from '../../redux/remoteData';
+import { selectIssueActionCreator } from '../../redux/callState';
+import { store } from '../../redux/store';
+import { getIssue } from '../shared/utils';
+import { remoteStateContext } from '../../contexts';
 
-interface RouteProps extends RouteComponentProps<{ id: string }> { }
+interface RouteProps {}
```
nickoneill: If route props is nothing, then remove it and don't inherit in `Props`

src/components/groups/GroupPage.tsx
```diff
@@ -1,36 +1,55 @@
...
+  appCacheContext,
+  groupStateContext,
+
+} from '../../contexts';
+
+interface RouteProps {
+  issueid: string;
 }
 
+// tslint:disable-next-line:no-bitwise
```
nickoneill: What's this about?

src/components/issues/MoreIssues.tsx
```diff
@@ -2,50 +2,135 @@ import * as React from 'react';
...
 
 interface Props {
   readonly activeIssues: Issue[];
   readonly inactiveIssues: Issue[];
-  readonly categoryMap: CategoryMap[];
   readonly completedIssueIds: string[];
-  readonly t: TranslationFunction;
-  readonly onSelectIssue: (issueId: string) => Function;  
+  // tslint:disable-next-line:no-any
+  readonly t: TranslationFunction|any;
```
nickoneill: `any`

src/components/shared/Promotion.tsx
```diff
@@ -6,7 +6,8 @@ import { Issue } from '../../common/model';
 
 interface Props {
   readonly currentIssue: Issue | null;
-  readonly t: TranslationFunction;
+  // tslint:disable-next-line:no-any
+  readonly t: TranslationFunction|any;
```
nickoneill: `any`

nickoneill (changes requested):
Looks like groups aren't loading - or they are loading but not displaying?

I noted a bunch of the translation `<any>` instances. There are a few that don't have this and aren't asking for it, so there should be a simple way to remove these.

We can leave the `dispatch<any>` in for now, I think we'll end up refactoring the dispatch calls out at a later time so don't worry about that for now.

Final thing is to make sure the tests are working again.

src/components/NotFoundPage.tsx
nabrahamson: I missed the translation in the footer component. It should be removed. The navigation component will be importing the i18next object and calling the translation function off of that like the other components are doing now. My next pushed changes will reflect this.

src/components/done/DonePage.tsx
nabrahamson: In order for RouteComponentProps to be extended, a type needs to be included. You'll see that RouteProps has been removed, but and empty object is being passed into the type parameter of RouteComponentProps now.

src/common/constants.ts
nabrahamson: That was my intent. I'll remove this constant as I've found that there are many places where static strings are embedded in code.

nickoneill:
Groups load now, but the sidebar doesn't reflect the current team or the use of custom issues if the team has them.

See the difference between https://5calls.org/team/nick and http://localhost:3000/team/nick

nabrahamson:
The problem with the sidebar was that the currentGroup wasn't being passed into the sidebarHeader. Should be working properly now.

nickoneill:
Sidebar is working great for groups!

nickoneill:
OK one other issue: if you navigate to a group page with custom issues, then click the header to get back home, the sidebar (header & issues) don't revert back to the main ones

nabrahamson:
I've added a prop to the layout page that specifies if group issues should
be shown in the sidebar or not. I've tested this on both the team/group and
team/group/issue pages.

nickoneill:
Looks good!

nickoneill (approved):
👍

PR #418, opened 2018-08-30
Profile Page

nickoneill:
Using the new profile API endpoint, we're able to display details of calls from the last 30 days along with total call stats. A solid starting point.

Use along with the `5calls-api` profile PR changes until it's merged.

Also removes onesignal, I don't think it's more useful than SMS for reminders on web.


src/components/profile/ProfilePage.tsx
```diff
@@ -1,22 +1,189 @@
...
+
+    this.state = { loadedProfile: false };
+  }
+
+  componentDidMount() {
+    queueUntilRehydration(() => {
+      getProfileInfo().then(profile => {
+        this.setState({ profile: profile, loadedProfile: true });
+      }).catch(error => {
+        // console.log('error getting profile', error);
```
nabrahamson, line 77: At some point we should be using an actual logger instead of letting things fail invisibly. Either storing things in a text based error log, or logging them to the database. (Preferrably the former than the latter).

nickoneill: Open to this. Can we manually log things to sentry? We already do automatic JS error catching there.

src/redux/remoteData/asyncActionCreator.ts
```diff
@@ -1,20 +1,22 @@
 import { Dispatch } from 'redux';
 import { ApiData, GroupIssues, IpInfoData, LocationFetchType,
   CountData } from './../../common/model';
-import { getAllIssues, getGroupIssues, getCountData, postBackfillOutcomes } from '../../services/apiServices';
+import { getAllIssues, getGroupIssues, getCountData,
```
nabrahamson: For purposes of making code reviews easier to see what's been added/removed from a list of imports, when importing multiple objects, they should be one per line like 

```
import {
  getAllIssues,
  getGroupIssues,
  getCountData, 
  ...
} from 'packageName';

nabrahamson (changes requested):

src/components/profile/ProfilePage.tsx
nabrahamson: I will have to look into this as I'm not as familiar with sentry as I should be.

src/redux/remoteData/asyncActionCreator.ts
nickoneill: I'll look to see if this is a tslint option

nickoneill: I'm actually not as much of a fan of this as I thought I would be. Doing this for every import makes for excessive numbers of lines that you have to scroll through before you reach actual code. There isn't an easy tslint for managing this style either.

A proposal for this: for imports that would extend beyond the line length limit, use the each-line style. Otherwise, use the normal style. I've updated the imports on this page to reflect this, but part of the problem here is that this file is really excessive. Too much stuff going on in one location.

nabrahamson (approved):

PR #423, opened 2018-09-17
Added a midterm challenge page

nickoneill:
Not much excitement here. New midterm challenge landing page, and home design to go along with it. Small API endpoint to get midterm challenge numbers from the API too.

Notably: using an event emitter to communicate across components via a event context. It would be nice to have some stronger types around this, I may look into something that is more typescript-focused rather than stringly-typed events names.


nabrahamson:
When pulling down the updated npm packages, I get a message stating that 5calls/react-components does not have a version available at 0.0.14

@nickoneill Because of this version inconsistency, your branch doesn't build on my machine.

src/components/MidtermsPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  week4actions() {
+    if (this.state.midtermStats && this.state.midtermStats.week4 > 0) {
+      return <span> ⋆ {this.state.midtermStats.week4} actions taken</span>;
+    }
+
+    return <></>;
+  }
+
```
nabrahamson, line 72: Did you mean to have weeks 5-8 included in this as well?

nickoneill:
OK, you should be good to go on this. I hate micro versioning components for minor changes.

nabrahamson:
@nickoneill Everything runs well after updating react-components. Looking at the event emitter pattern, This seems like a straight forward way to pass the events around in a DRY way with minimal code.

nabrahamson (changes requested):

src/components/MidtermsPage.tsx
nickoneill: Yeah, they're revealed over time though. I'll add the rest of them in today with another small change I want to make.

nickoneill:
Fixed these items, merging in.

PR #428, opened 2018-10-29
Run End-to-End Tests on CircleCI (using Browserstack)

### LINKED ISSUE #291: Configure E2E tests to run in CircleCI and SauceLabs
Now that we have the suite of end-to-end tests in #252 merged into master, we need to configure CircleCI and SauceLabs to run these tests. Rob (@Mr0grog) has kindly agreed to lead this effort.

Mr0grog:
This should resolve #291 by running most of the end-to-end tests with Browserstack in our normal CircleCI builds. **Note: someone who has the right Browserstack credentials will need to set `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY` in the CircleCI console in order for this to work.**

The vast majority of the heavy lifting needed back when we first looked at this is now already done by Testcafe (and the testcafe-browser-provider-browserstack plugin). Setting up our dev server to run in the background is fairly straightforward. I actually spent most of the time here dealing with other intermittent failures that made the end-to-end tests unreliable in the first place:

- Weird, intermittent failures (about 1/3 of the time) caused by the `testcafe-react-selectors` package (see https://github.com/5calls/5calls/commit/31e8920661aaaa6797771f1fe22e57e024589f75#diff-9d5b2b71594d1c8b15f7e2762d4d99a3R30). To solve this, I wound up commenting out these specific React-based selectors in favor of using simple HTML-based selectors. I definitely didn’t have time to dig into what kind of weird race condition is really happening here, but this change made tests immediately more reliable. I left the old code present and commented out as I assume it would be preferable to switch back to if we can solve the problem.

- Failures caused by changing data from the API. Two of the tests are looking for content that will only be present for some types of issues and, since the API is regularly changing what data you get (and even sending different data depending on where you are running the test from), it’s pretty hard to reliably know whether the test will receive the data it needs to succeed. What we really need here is some kind of fixture, mock API, or proxy API server that sends pre-recorded responses. That’s a big change, though, so I’m not attempting it in this PR. Instead, I’ve simply skipped those tests when running on CI.

I also noticed that CircleCI is configured to run the tests on Node.js v7, which is an unstable release line. (This is the reason I had to [use `process.exit(code)` instead of `process.exitCode = code`](https://github.com/5calls/5calls/commit/befe365d7985291475220b611dae825c0b015e91#diff-b9c2b69f1fd91e39237ff14152db7467R50).) I’m happy to add a commit here or to create a separate PR to upgrade to Node.js v8. Let me know whichever is preferable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
             wget --retry-connrefused --no-check-certificate -T 60 localhost:3000
 
             # Actually do the tests
-            yarn web-tests:browserstack
+            yarn run web-tests:browserstack
             EXIT_CODE=$?
 
             # Stop server
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,3 +29,20 @@ jobs:
         
       # run tests!
       - run: yarn test
+
+      - run:
+          name: "End-to-End Tests"
+          command: |
+            # Start test server
+            yarn start &
+            SERVER_PID=$!
+            # ...and wait for it to be ready
+            wget --retry-connrefused --no-check-certificate -T 60 localhost:3000
+
+            # Actually do the tests
+            yarn web-tests:browserstack
+            EXIT_CODE=$?
+
+            # Stop server
+            kill -9 $SERVER_PID
+            exit $EXIT_CODE
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image: circleci/node:7.10
+      - image: circleci/node:8.12
 
 
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,9 @@ jobs:
             yarn start &
             SERVER_PID=$!
             # ...and wait for it to be ready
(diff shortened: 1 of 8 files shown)


Mr0grog:
FYI: you should be able to see the tests working here because this PR is coming from my fork, which is running on my CircleCI account, which has credentials for a free Browserstack account plugged into it. Once this is merged in, you’ll have to make sure the right credentials are added to 5calls’s CircleCI config.

nickoneill:
Thanks, this is awesome!

Appreciate the work on react selectors, @nabrahamson was having trouble with them as well. I think he filed an issue against that repo for those issues at some point...

I didn't realize we were specifying nodev7, let's definitely switch to v8, I'll leave it to you if you want to make a new PR or not. I'll add those browserstack keys in to Circle and check things out in detail later today.

Mr0grog:
Since it sounds like upgrading Node on Circle is non-controversial, I went ahead and just added a commit here :)

nickoneill (approved):
Looks great 👍

nickoneill:
@Mr0grog do you mind creating a new issue with some info about making the skipped tests `nth()` choice stable? Want to keep these suggested fixes in mind going forward.

Mr0grog:
Oops, totally missed your comment, @nickoneill! Just added an issue in #431.

PR #432, opened 2018-11-18
add rel=noopener

aholachek:
[Links that use target="_blank" should all use rel="noopener" for security and performance reasons.](https://developers.google.com/web/tools/lighthouse/audits/noopener)

There are some links in the `Promotion` component that I didn't alter since they are doing something more involved with share widgets that I didn't fully understand.

### DIFF
--- a/src/components/donation/Donation.tsx
+++ b/src/components/donation/Donation.tsx
@@ -19,19 +19,19 @@ export class Donation extends React.Component<Props, State> {
               Get Involved:
             </p>
             {/* <div className="logo__header__donatebutton">
-              <a href={`${Constants.WRITER_APPLY_URL}`} target="_blank">Apply</a>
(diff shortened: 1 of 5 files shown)


nickoneill:
This is great @aholachek, thanks for putting this together! I had no idea these kinds of rules existed for opening new windows.

nickoneill:
Ah - that test failure is the linter breaking on some long lines. Can you take a look?

nickoneill:
The snapshot tests just need to be updated for these new parameters, if you run them locally with `yarn test`, it'll give you a chance to update them after they fail.

Rather than ignoring the long line, can we break it into multiple lines so the code maintains readability?

nickoneill (approved):

nickoneill:
Awesome, thanks for fixing those tests! Will merge in later today.

Issue #424, opened 2018-09-23
Mobile doesn't scroll to new content


nickoneill:
> On my phone, just clicked http://5calls.org/midterms then the first issue. Nothing happened. Hmm… second issue, nothing happened. Then realized that I needed to scroll down to see the content that had changes

MonaTem:
Nick should I look at this?  What type of phone do you have..make and model?

nickoneill:
@MonaTem I believe another volunteer is going to take a look at this one. If not I'll ping you.

MonaTem:
Ok…

I’m trying to get on the Slack channel, but so far no reply from hello@5calls.org <mailto:hello@5calls.org>

nickoneill:
Sorry about that, just added you now.

MonaTem:
Thanks…

PR #433, opened 2018-11-18
scroll to heading on mobile

aholachek:
Hi, I saw [this issue](https://github.com/5calls/5calls/issues/424) had not been responded to for a while so I made a pr.

In this implementation, Chrome and Firefox will smooth scroll to the heading on screen sizes less than 768, while Safari and Edge will jump directly to the heading without the scroll animation.

I am relying on there being only 1 `h1` as a child of the `Call` component, but that is required by semantic HTML  so I think it's an ok tradeoff for not having to forward refs on the `CallHeader` component.

The scroll is currently also called when the component mounts, not sure if that was desired, if not I can remove. To be safe, I am triggering the scroll on all `componentDidUpdate`s, but it's possible it should only be called when certain props change, I wasn't sure.

Additionally, this should work in IE11, but I wasn't able to test it because it was [erroring on this line locally](https://github.com/5calls/5calls/blob/master/src/components/layout/Layout.tsx#L21) If you don't support IE11 than I can simplify the logic for the `crossBrowserScrollY ` variable.

I haven't done a lot of Typescript before, I would be happy to make any changes that are required.

### DIFF
--- a/src/components/call/Call.tsx
+++ b/src/components/call/Call.tsx
@@ -44,7 +44,7 @@ export class Call extends React.Component<Props, State> {
       return;
     }
 
-    const pageTitle = this.containerRef.current.querySelector('h1');
+    const pageTitle = this.containerRef.current.querySelector("[data-scroll-anchor]");
     if (!pageTitle) {
       return;
     }
@@ -96,14 +96,15 @@ export class Call extends React.Component<Props, State> {
   }
 
   componentDidMount() {
-    this.scrollToTitleOnMobile();
+    // wait a tick to ensure rendering of children is complete
+    setTimeout(this.scrollToTitleOnMobile, 1);
   }
 
   componentDidUpdate(prevProps: Props) {
     if (!isEqual(prevProps, this.props)) {
       this.setState(this.setStateFromProps(this.props));
+      this.scrollToTitleOnMobile();
     }
-    this.scrollToTitleOnMobile();
   }
 
   // this should obviously be somewhere on issue but as an interface and not a class I don't know where...
--- a/src/components/call/Call.tsx
+++ b/src/components/call/Call.tsx
@@ -22,11 +22,48 @@ export interface State {
   numberContactsLeft: number;
 }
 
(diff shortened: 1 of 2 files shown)


nickoneill:
Thanks @aholachek! Will review in more detail shortly.

IE 11 isn't an insignificant amount of our traffic, so I've opened #434 to track that issue. I appreciate surfacing that error!

nickoneill:
Interesting use of scrollBehavior, I wasn't aware of it. Support is *rough* right now: https://caniuse.com/#feat=css-scroll-behavior, which is too bad because the effect really helps explain what's happening visually.

I noticed that after all the contacts have finished and we navigate to the **done** page, the window scrolls to the top of the page, which is a different behavior than it does after incrementing contacts. This is a different component than `Call` which might be conflicting with the `containerRef` functionality?

Do you think there's a way to make it more intentional where the page should scroll to and more modular for different pages? Off the top of my head: maybe inserting some non-rendering element (or attribute) in the jsx to indicate a scroll target, then calling something in `componentDidMount` and `componentDidUpdate` like you have now that finds and navigates to it.

My concern is two-fold: one that it would be easy to move/misplace an h1 or some other target that we intended on using for scrolling behavior, and two that this is something we do really want to use in various components (like the done page as noted above) and something more modular would help implementing that in other places.

aholachek:
Sure, what if I put a data-attribute on the heading like `<h1 data-scroll-to>blah</h1>`? 
I think I need something in the DOM to know what coordinates to render to.

If you'd prefer smooth scrolling for all browsers, I could use something like [this small library](https://github.com/alicelieutier/smoothScroll/blob/master/smoothscroll.js). I wasn't sure what the policy was for adding new libraries, so I used the progressive enhancement approach. What do you think?

nickoneill:
Yeah! `data-scroll-to` is definitely along the lines of what makes sense here.

Don't worry about the smooth scrolling everywhere to start, let's start simple and move forward from there.

Issue #434, opened 2018-11-19
Check tests on IE 11


nickoneill:
From this PR: https://github.com/5calls/5calls/pull/433

It's noted that IE 11 fails on one of our usages of `find`, which we could probably replace easily with lodash (or polyfill?).

a) Do our tests not catch IE 11 failures right now?
b) Decide on a fix path and fix this issue.

nickoneill:
Tests are in #435 to confirm any fixes someone may make against this issue

nickoneill:
Fixed in #456

PR #435, opened 2018-11-20
Added ie 11 and edge to our testing matrix

nickoneill:
Working on getting #434 to fail

### DIFF
--- a/runner.js
+++ b/runner.js
@@ -3,7 +3,7 @@ const createTestCafe = require('testcafe');
 let testcafe = null;
 
 const browsers = [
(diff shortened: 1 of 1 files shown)


nickoneill:
OK.

IE 11 failing on `find` as expected now, and Edge failing in ReactSelector which is probably related to the sketchy failures that @Mr0grog was experiencing earlier.

I'll use this PR to fix the `find` issue and move away from ReactSelector in tests.

nickoneill:
Closing in favor of #500

PR #438, opened 2018-11-21
add precommit hook for tslint fix

aholachek:
Hi, in preparation for possibly adding prettier later on, this just sets up a precommit hook that runs tslint --fix, and fails if there are any tslint errors remaining after the fix step.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,13 @@ that you're going to be working on it and start coding!
 Once a feature has been reviewed and approved for implementation, please follow
 the following guidelines for submitting your Pull Request (PR):
 
-1. Be sure your PR branch is rebased if it becomes out-of-date with the 5 Calls
+1. Each time you commit, an `tslint --fix` command will automatically be run and
+   attempt to fix any lint errors found in the code. If errors remain that cannot
+   be automatically corrected, the commit will fail with a message indicating
+   where the remaining lint errors are. You'll have to fix them before the commit will happen.
+   Alternatively, if you'd like to bypass this checking for some reason, you can run
+   `git commit -m [commit message] --no-verify`.
+2. Be sure your PR branch is rebased if it becomes out-of-date with the 5 Calls
    repo's master branch.
 3. A unit or end-to-end test (or tests) should be included with the Pull Request
    covering the changes you've made. If tests are not possible, an explanation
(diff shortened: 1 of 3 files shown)


aholachek:
I guess I ran into [this issue](https://github.com/yarnpkg/yarn/issues/6440) with the yarn.lock file, I'll try to fix it later today

nickoneill:
👍 I'll give this a shot locally a bit later today.

I'm not real well read on the details of yarn updates (other than reading that thread). Is there something more general that's keeping folks from upgrading to yarn 1.10.x across different projects?

aholachek:
I think it's ready to go now. (I upgraded yarn, I'm using brew and I learned I can just do `brew switch yarn [old version]` for other projects that require an older yarn version.)

To test the precommit hook, I added the line `console.log("foo")` in index.tsx. As expected, when I tried to commit the change, two things happened:
1. the task replaced the line with `console.log('foo');`
2.  the commit failed because `console.log` is not allowed by tslint, and that's not something the fix command can auto fix.

nickoneill:
I tried the same thing with this branch and it let a console commit through fine.

What is `husky` and are people expected to have it installed to use this? I don't think I have it.

Also, as we wrap this up for merging, can we make sure this is documented in the readme / contributing section so folks know what's up when they're committing for the first time.

aholachek:
Ah husky should be getting installed in the `devDependencies`. `Husky` plus `lint-staged` is a workflow suggested by e.g. [prettier](https://prettier.io/docs/en/precommit.html) to do work in the pre-commit stage and optionally cancel the commit.
This is a dumb question, but Is it possible you didn't run yarn install to add husky? If not, I'm not sure why it let you commit the console statement. I'll check it out now, and I'll add something to the Readme as well as you suggested.

nickoneill:
👍 Yep, I didn't re-run yarn after pulling the branch. Looks like it works well now!

Using `tslint --fix` is different than the code organizing that `prettier` does, right? If not, I think we should do a pass on the entire codebase with `prettier` before merging so new commits from folks don't have a bunch of unexpected changes.

aholachek:
They are definitely different, I was planning on making a second pr to add prettier once this got merged. The final precommit task should run `prettier` and then `tslint --fix` in that order, I think.

I will definitely format all the files in that next pr, so that people going forward don't have diff issues.

nickoneill:
Cool, we're definitely ready to bring this in then, thanks for working on it!

nickoneill (approved):

PR #439, opened 2018-11-22
An update for test configuration, since mixpanel broke it

nickoneill:


nickoneill:
Did this in #440 differently

PR #441, opened 2018-11-25
add prettier

aholachek:
Hi, here's the codebase run through prettier, and with prettier added as a precommit hook step.

Prettier enforces a max 80 char line width by default so I went ahead and removed the conflicting tslint line length rule. That means that we shouldn't have to use the `tslint-disable-line:max-width` anymore since prettier will fold anything it can, and anything that can't be folded can be left alone.

Obviously this diff is really big, I've been trying to QA the changes and everything seems ok but it would be great to get more eyes on it.

One issue is that commits take slightly longer now that the checks are being run, but I think that's an ok tradeoff for the increased standardization/convenience of auto formatting.  The [vscode prettier plugin](https://github.com/prettier/prettier-vscode#vscode-eslint-and-tslint-integration) has a tslint integration that should allow autoformat on save as well.


nickoneill:
I mentioned this on slack but for clarity: can we rebase this so the new netlify config gets built and we can use the preview link to test to see if there are any obvious issues?

nickoneill:
This looks great! A couple things I noticed:

* The deploy preview works now, and looks good. I tested a few different scenarios as well as made sure submitting still hit the database with results and everything looked great.
* Text that wraps looks good - there's some weird spacing stuff that prettier does when it wraps around tags inside text but it seems like it knows what its doing.
* The upgrade to run-all actually drops support for `ps-tree` in favor of `pidtree` (`ps-tree` had that nasty cryptocurrency malware in it - not the version we use but happy to get away from it in general)

I'm really happy with how this turned out, thank you for putting the effort into getting this together.

nickoneill (approved):

PR #444, opened 2018-11-29
Simple typography changes to enhance site readability

aholachek:
This is probably not quite ready to merge, it could use some more QA first.

I'm happy to make any further changes required, or undo anything. 

My auto formatter introduced a few unnecessary changes which I haven't removed, let me know if it's too much of a distraction and I can try to undo them!

localhost_3001_issue_anti-immigrant-chad-readler-judge


aholachek:
Oh actually now that I'm thinking about it I should have run through the scss files through prettier as well, I can do that in a separate branch

nickoneill:
Conflicts are from merging the scss formatting, it should be OK with a rebase.

The general direction here is good, but there's a lot of spacing / sizing that could use a polish so everything fits nicely. I'll do a full pass and post some screens tomorrow - thanks for putting this together!

nickoneill:
Sorry for the wait here @aholachek, this is an excellent change and improves readability tremendously. I've got a couple questions on browser treatment and some small tweaks to request.

Since we're providing the font Roboto from google fonts, shouldn't we have the same set of weights?
we can drop the italics on this as well. Safari vs Chrome on mac:
screen shot 2018-12-10 at 5 18 56 pm
screen shot 2018-12-10 at 5 19 10 pm

Some font size/weight adjustments on location:
screen shot 2018-12-10 at 5 25 46 pm
(I'm editing live in chrome to get a sense of size and screenshot, I know we use px and rem both in the app, I don't have great opinions on how we should do this better)

Adjustments on result buttons:
screen shot 2018-12-10 at 5 29 51 pm

"Get Involved" bar needs some vertical centering of buttons and text, there's uneven space above and below the buttons and the text doesn't feel vertically centered within the buttons (no image edits on this one)
screen shot 2018-12-10 at 5 32 07 pm

One last thing I want to do after these changes: we should take a look on whatever mobile browsers we have on hand as well as a few other OSes / browsers via browserstack to make sure everything is looking good there too.

Let's make sure to do a pass on at least these browsers:
 - [ ] Latest iOS
 - [ ] Latest 2 versions of Android Chrome
 - [ ] Firefox, Mac and Windows
 - [ ] Edge, Windows
 - [ ] Chrome, Mac and Windows

aholachek:
Hi @nickoneill thank you for the detailed review! I'll plan on working on this over the weekend.
I'll use browserstack to test the targets you listed (I can test on an actual iphone) if that's sufficient.

nickoneill:
Thanks @aholachek! I will happily help test on multiple browsers when you're all set with the changes.

nickoneill:
I just pushed a large change to master, I think it makes sense for me to tackle the conflicts here so I'll take a look later today and then we can get this merged in!

nickoneill (approved):

nickoneill:
Merged! Thank you @aholachek!

PR #456, opened 2019-01-08
Added a polyfill for find in ie11

nickoneill:
Should fix the tests in #435 as well


nickoneill:
Confirmed in browserstack that this fixes the problem in ie11

Issue #455, opened 2019-01-08
Upgrade to webpack 4


nickoneill:
We're a bit behind in webpack versions, it would be nice to be on a modern version of webpack but the dev/prod configs will probably need to be looked at for errors / defaults in the upgrade from 2.6.0->4

kamoh:
I was going to take a stab at this, but looks like you did it @nickoneill - close this out?

ref: https://github.com/5calls/5calls/commit/0826d020f839eb3298be9cbacf03e8b2e912e86c

nickoneill:
Yep, this is now resolved. If you want to take a look at #495 and reducing our overall JS size, any of the specific items I mentioned in the comments are good spots to dive in, specifically looking at reducing / replacing `react-phone-number-input` might be a good starting place.

PR #459, opened 2019-01-09
Don't generate sourcemaps for css

nickoneill:
They're getting mixed with the js ones because I dunno, maybe webpack is old?


nickoneill:
Ugh, this is not simple anymore. Maybe it will work correctly if we update to webpack 4? #455

nickoneill:
Resolved this by updating webpack in #455

Issue #465, opened 2019-01-22
Unordered lists in issues should have a smaller font size


nickoneill:
Font size is too small on the unordered lists, it should be a similar size and line height as the rest of the text.

<img width="806" alt="screen shot 2019-01-21 at 8 13 05 pm" src="https://user-images.githubusercontent.com/49688/51512102-14679a00-1db9-11e9-9de4-360c3a44a3d8.png">

nickoneill:
@aholachek do you have some time to take a look at this styling issue?

nickoneill:
Fixed in #465 (thanks @aholachek!)

PR #466, opened 2019-01-22
#465 call reason ol and ul sizing and line height updated

aholachek:
addresses #465 

I'm keeping the edits scoped to `.call__reason` since I didn't want to hunt down every `ol` in the site and make sure that they were supposed to be 18px.

#### After the edits:
screen shot 2019-01-22 at 8 43 53 am

### DIFF
--- a/src/components/shared/scss/_call.scss
+++ b/src/components/shared/scss/_call.scss
@@ -245,8 +245,14 @@
 }
 
 .call__reason {
-  p {
+  p,
+  ol,
+  ul {
     line-height: 1.6;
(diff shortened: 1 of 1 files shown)


aholachek:
Oh, it might have just been me, but I needed to comment out everywhere the `bugsnag` client was used in order to get the site running locally

nickoneill:
Thanks @aholachek!

What was the problem bugsnag was giving you? It should disable itself in dev.

nickoneill (approved):

PR #468, opened 2019-01-23
remove old donation banner, slight redesign of header, rename confusi…

aholachek:
…ng breakpoint mixin to max-width-bp, add a more traditional breakpoint mixin

I'm of course happy to make any changes that are required, or play around with styles as necessary.

I'd love to get my hands on the @5calls/Login component to fix some stuff (I personally would put a more obvious login icon rather than the 5 stars logo, and I'd put it on the left rather than the right), and the markup for the component could be improved in any case) but I'm not sure where it lives and I'm too lazy to find it!

I think we really need  a better, more consistent, set of modern looking buttons. I made a custom style for the "donate" button on hover but it's not quite consistent with the rest of the app. The button that goes from light blue to red on hover on the "change location" button looks pretty dated to my eyes.

Maybe that's another high-impact small task -- to make primary, secondary, and outline button components that we can use throughout the app.

On the topic of components, I personally would prefer to organize the sass on a more component basis with something like css modules or styled-components, I think it would help keep the project organized and usable for various devs (since it's less fraught when you want to delete something or change it around, since you know exactly where it's used).

Before on mobile: 

screen shot 2019-01-22 at 9 08 36 pm

After:
screen shot 2019-01-22 at 9 08 17 pm

Full screen:

screen shot 2019-01-22 at 9 08 26 pm


aholachek:
I'd also like to figure out what's going on with the flash of somewhat un-styled content when the page first loads, and try to minimize that jump, although that was also a problem before this pr I think.

nickoneill:
You are amazing @aholachek. The flash of content has been bothering me for a long and I assumed it was part of some of our i18n work, but I recently made a test branch removing the i18n stuff (I am normally for i18n but literally no one uses it on the site, and it likes to break and show untranslated content more often than I would like) to see if it would fix it and it did not.

It might be related to a lack of css inlining, we're on a version of webpack that is a bit old (#455) which has been showing a few signs of weirdness here and there. Feel free to investigate but probably worth a new branch because it might get complicated.

aholachek:
@nickoneill I think the flash of weirdness is due to the fact that there's a bunch of markup in the index.html that's getting rendered before the JS loads: https://github.com/5calls/5calls/blob/master/public/index.html

I personally would remove it since it's 1) not consistent with the styles or content of the js-rendered app and 2) pretty jarring, but someone could also fix it to look better if it's absolutely necessary.

nickoneill:
I wonder what the experience would be like on load for a page without any markup? Showing a blank page for a similar amount of time is small improvement in my eye, but if removing the markup shortens the time until we see the real React content, that'd be a good move.

aholachek:
I made [another pr](https://github.com/5calls/5calls/pull/469) for proper pre-rendering, if you could take a look.

nickoneill (approved):

nickoneill:
This looks great, thank you @aholachek! I've moved the button suggestion to #470

PR #469, opened 2019-01-23
server rendering mostly works

aholachek:
This is **not ready** to merge, it's just a proof of concept of server rendering using [this library](https://github.com/stereobooster/react-snap).

To test it out: [view the deploy preview](https://deploy-preview-469--5calls.netlify.com) with js turned off

There are still some rough edges (on the call page the ajax request clearly needs to be fixed) but in general it seems to work.

Caveats: if code splitting is ever used, I think this will break.


aholachek:
Also, if a new call page is added and the build step isn't re-run, there would be no prerendered HTML so I think it would just work like a normal SPA, but that would have to be tested.

nickoneill:
That does feel a lot snappier for a first load, and absolutely fixes the flash of content issue! I've been wanting to do some sort of server-side rendering for a while but not really sure where to start. This is... actually pretty amazing and I'm not sure I understand the implication yet.

I'm not 100% clear on the interaction between netlify and react-snap. We have a line in our netlify config that basically says "this is a single page app, all requests to be made to index.html" but clearly the call pages are returning pre-rendered content since they're not fetching the issues like they do in the SPA. Can we find any more details verifying that netlify supports this behavior? Or at least it's been tried before?

Not too worried about code splitting, I think we're a small enough site that we should simply be mindful of our deployed js size in total (about 300kb currently).

For tracking, two todo items:
- [x] Keep integrity hashes in the yarn lock file
- [x] Fix issue + contact fetching when starting up in a call page

aholachek:
I can't find the relevant netlify docs, but I *assume* what is happening is netlify first looks for a matching route on the filesystem, and if it doesn't find one, defaults to the 200.html page. That's how it works on surge.sh, which I'm more familiar with. Since the pages are now statically built, netlify just serves them like a classic server would.

nickoneill:
Oh, you can actually see all the caching happen in snap in the netlify deploy log. I am really impressed, it seems like this is a pretty cool and simple solution. Snippet below:

```
...normal build stuff above...
6:17:36 PM: The build folder is ready to be deployed.
6:17:36 PM: You may serve it with a static server:
6:17:36 PM:   yarn global add serve
6:17:36 PM:   serve -s build
6:17:37 PM: $ react-snap
6:17:39 PM: ️️️💬  console.log at /: [bugsnag] Loaded!
6:17:39 PM: ️️️💬  console.log at /: [bugsnag] Session not sent due to releaseStage/notifyReleaseStages configuration
6:17:39 PM: ️️️💬  console.log at /: i18next::translator: missingKey   issues.viewAllActiveIssues issues.viewAllActiveIssues
6:17:39 PM: ️️️💬  console.log at /: starting to listen for event
6:17:39 PM: ️️️💬  console.log at /: current listeners login JSHandle@array
6:17:39 PM: ️️️💬  console.log at /: i18next::backendConnector: loaded namespace  for language en JSHandle@object
6:17:40 PM: ️️️💬  console.log at /: i18next::backendConnector: loading namespace  for language en-US failed failed parsing /locales/en-US.json to json
6:17:40 PM: ️️️💬  console.log at /: i18next: languageChanged en-US
6:17:40 PM: ️️️💬  console.log at /: i18next: initialized JSHandle@object
6:17:42 PM: ✅  crawled 1 out of 16 (/)
6:17:42 PM: ️️️💬  console.log at /issue/support-legislation-to-protect-transgender-troops: [bugsnag] Loaded!
6:17:42 PM: ️️️💬  console.log at /issue/support-legislation-to-protect-transgender-troops: [bugsnag] Session not sent due to releaseStage/notifyReleaseStages configuration
...many page crawls and i18n errors later...
6:18:02 PM: ✅  crawled 38 out of 40 (/issue/demand-the-closure-and-investigation-of-child-migrant-prisons)
6:18:02 PM: ✅  crawled 39 out of 40 (/issue/protect-mueller-and-the-russia-investigation-)
6:18:03 PM: ✅  crawled 40 out of 40 (/issue/demand-acting-ag-matthew-whitaker-recuse-himself-from-mueller-investigation)
6:18:03 PM: Done in 56.08s.
```

Do you want to take a look at how we can fetch the issues list in a way that works for both SPA and pre-render? If not, I'm happy to check into it. Issues are fetched in a `startup` function that runs after the redux store is finished starting, I'd be interested in understanding if this is bypassed with pre-rendering or if it runs but in a different order than we expected? Is the pre-rendering caching a page that says "no calls to make, this issue can't be found"? Because ideally it would cache the page with the actual issue text on it.

aholachek:
Re: the call page issue, the server render is actually working fine, it's the JS that is run on page startup that fails (you can check by disabling JS and loading a call page, it looks fine).

I've pinpointed the issue to [the line here](https://github.com/aholachek/5calls/blob/master/src/components/call/CallPage.tsx#L87) that on startup is dispatching an empty string for the `CURRENT_ISSUE_SELECTED` action.  (You can see it in the redux devtools). Maybe the prop coming from react-router is wrong when the app starts up or something. However, it's tough to debug right now because I would need to debug the built files and the sourcemaps are not working for me in chrome or firefox, so I can't put a debugger into the code to see what's happening. I would be happy for you to take a look.

nickoneill:
Yeah, the sourcemap issue is on my list to check out too. I'm not sure when it started but for some reason webpack is pointing the current sourcemap at the css bundle 🤨

I will take a look at getting those requests working for both SPA and snap-rendered pages, thanks for being open to collaborative PRs!

aholachek:
Please feel free to commit to any branch I work on, I have no problem with that!

Update: I refactored the CallPage.tsx component to remove some of the complications related to the derived state from props, and the problem is now fixed. However, there is still a lesser problem-- a flash during data fetching where it goes from "nicely rendered page" --> "no calls available" --> nicely rendered page, and I'm not sure how to get rid of it. I wonder if you could take a look or point me in the right direction?

I know we talked about this a couple months ago, but having taken a look at and refactored CallPage it seems like the view components in general could really use refactoring in two ways:
1. The derived state stuff hugely complicates things for no reason (that I can tell at least) I was able to remove 30 lines from callpage with (I think) the exact same functionality.
2. The use of context wrappers instead of the traditional redux/react-redux connected components makes things much more complex (to my eyes, at least) as well. I wonder if I'm missing some rationale behind it? According to git commits it was called something like "react 16 context refactor". Simplifying data flow to use idiomatic redux patterns would mean less code and also much more clear data flow, I think.

(I'm on vacation this week which is why I'm spending a lot of time on this repo, I will be less noisy starting next week!)

nickoneill:
I will take a look at the flash of "no calls", but it'll take me a bit as I want to dig into how snap rehydrates the existing html on the page.

As for the context wrappers: we replaced a lot of the traditional `connect()` redux hookups because there was *so much* prop drilling (props passed down through 3-4 components just to get to the bottom one) going on that it was hard to tell what was used where. Context might be noisier per-component but it simplifies intermediate components that don't need to know about `userStats` when its used below them. Open to simpler solutions if they exist!

aholachek:
With correct use of react-redux container components there should be no prop-drilling (which I also hate!). [Each component that needs to access redux can be wrapped with its own connected higher-order component (this is better for performance too)](https://redux.js.org/faq/react-redux#should-i-only-connect-my-top-component-or-can-i-connect-multiple-components-in-my-tree)

nickoneill:
This looks really solid! I love the refactor of the callpage props, much simpler. Do you want to take a shot at rebasing this on top of master and then we'll roll it out for 30% of users?

aholachek:
There's still an ugly flickering when the data is fetched on the client side, I'd like to figure out how to implement [redux hydration](https://github.com/stereobooster/react-snap#redux) to avoid that, it looks like it should be pretty straightforward

aholachek:
I can't get react-snap's redux caching to work (there's an error being thrown in a component somewhere when the react tree mounts) and unfortunately without sourcemaps it's hard to debug. 

I think the current flickering is still better than what we have now, so it's up to you if you think it's ok to merge as is.

nickoneill:
It's definitely better than what we have now, and I think we can move forward on deploying this without redux hydration.

I'll create a new issue tracking improvements in hydration, and put the webpack / sourcemaps issue higher up in my queue to do.

nickoneill:
Merging this into staging

Issue #470, opened 2019-01-24
Improve button styles a la donate button


nickoneill:
From @aholachek:

> I think we really need a better, more consistent, set of modern looking buttons. I made a custom style for the "donate" button on hover but it's not quite consistent with the rest of the app. The button that goes from light blue to red on hover on the "change location" button looks pretty dated to my eyes.

> Maybe that's another high-impact small task -- to make primary, secondary, and outline button components that we can use throughout the app.

Fully agree. The new donate button on the header feels like a nice improvement, let's base our work on that as a starting point.

aholachek:
I briefly went down a rabbit hole trying to get optional css modules set up for this before realizing that it's not really necessary, as much as I'd like to move over to some sort of css-in-js solution, I'll just do some edits of the current css for now for the buttons.

It did get me wondering, though, how you would feel about trying to migrate the code over to a non-ejected create-react-app v2? That handles typescript, sass, and css modules out of the box, and I bet would fix that source map issue as well. I don't know enough about the project to know if there are serious reasons that it's not feasible, but thought I might bring it up before spending more time thinking about the current setup.

PR #471, opened 2019-01-25
Buttons

aholachek:
Addresses [#470](https://github.com/5calls/5calls/issues/470)

This ended up being a bigger pr than I intended because I bundled in a few other UI tweaks as well.
Please take a look and let me know if you'd like anything to be changed back/ changed in some other way.

My goal here was to only use the new "secondary" button with a dark blue background to denote important actions (which is why I also changed the footer background -- we should ideally use bright backgrounds only on actionable components.)

I'm not sure if Roboto is just not rendering nicely on my old macbook air, but I think the typography on the buttons could probably be improved further as well.

### DIFF
--- a/src/components/call/Outcomes.tsx
+++ b/src/components/call/Outcomes.tsx
@@ -109,6 +109,7 @@ class Outcomes extends React.Component<
             <div className="call__outcomes__items">
               {this.props.currentIssue.outcomeModels.map((outcome, index) => (
                 <button
+                  className="btn btn-secondary"
                   key={index}
                   onClick={e => this.dispatchOutcome(e, outcome.label)}
                 >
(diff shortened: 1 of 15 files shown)


nickoneill:
Huge improvement visually, I think.

I like most of the treatments for the font + buttons, except this one:
screen shot 2019-01-25 at 11 10 19 am

What do you think about this change for the style:
screen shot 2019-01-25 at 11 14 20 am

2px border, 16px font size, stay with 600 weight, center the text a bit better with `padding: .5rem 1rem .4rem 1rem`

aholachek:
I'm not sure why the tests are failing -- I think it's possible it's something that is not related to this pr

nickoneill:
Yep, looks like it just needed to be retested again. Will merge when it completes this time around.

nickoneill (approved):

PR #490, opened 2019-02-15
Adopt our previous variable name to use field offices

nickoneill:
Maybe this got missed in our prettier rework? The name is not cool by tslint standards, but it's in use by all clients so we're just going to make an exception for it at the moment. The field office design still looks good.

### DIFF
--- a/src/common/models/contact.ts
+++ b/src/common/models/contact.ts
@@ -9,7 +9,8 @@ export class Contact {
   state: string;
   reason: string;
   area?: string;
-  fieldOffices?: FieldOffice[];
+  // tslint:disable-next-line:variable-name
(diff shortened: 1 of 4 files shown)


nickoneill:
Finally finished the data side of the field office review. Just rebased this, hope to push it out with the data soonish.

aholachek (approved):
LGTM! Just so I understand, did prettier somehow convert the `field_offices` field to `fieldOffices`?  I don't think it's supposed to mutate variable names like that since it should be concerned only with formatting.

nickoneill:
Not sure... the json name from the API didn't change, so I'm not quite sure where it went wrong.

Issue #484, opened 2019-02-10
Investigate proper use of react-redux connect


nickoneill:
Inspired by: https://github.com/5calls/5calls/pull/469#issuecomment-457369221

We originally used traditional react-redux `connect()` to add props to our components, but got caught up in a lot of prop drilling to deliver props to deeply nested components. This was "simplified" with contexts for different pieces of state, but that can still be noisy.

Apparently there is a better way to do this with out prop drilling! Before fully diving in, let's try this approach out with a single page / set of components and see how it looks.

nickoneill:
@aholachek feel free to unassign yourself if you're not interested in this. It's not super pressing but I wanted to capture it separate from the react-snap PR.

aholachek:
I would love to do this! I'll plan on making the demo next weekend.

nickoneill:
Thanks!

PR #491, opened 2019-02-17
Simplify state

aholachek:
Here's a proof of concept for #484, refactoring the Callpage components. The goal is to make the data flow clearer and more concise to make it easier to update these pages in the future.

By connecting each component that needs redux data directly to redux, and only passing in the data that the component actually needs rather than an entire big object directly from the store, we can make the data requirements of each component more clear. Ultimately we would completely get rid of custom context, (other than for the event emitter which could probably just be refactored to use a thunk), resulting in less code.

You can see that, instead of adding prop drilling, it actually gets rid of it in a few places.

There are a few snapshot tests and possibly others that are still failing, if you like this direction I'll fix them when I update this pr.


src/components/call/index.tsx
```diff
@@ -108,23 +106,8 @@ class CallPageView extends React.Component<Props, State> {
   }
 }
 
-export const CallPageWithRouter = withRouter(CallPageView);
```
aholachek, line 111: withRouter shouldn't be necessary since this is a top-level route component

src/components/call/Call.tsx
```diff
@@ -87,62 +86,51 @@ export default class Call extends React.Component<Props, State> {
...
+            </eventContext.Consumer>
           </>
         )}
       </section>
     );
   }
 }
+
+const mapStateToProps = (state: ApplicationState) => ({
+  contactIndexes: state.callState.contactIndexes
```
nickoneill, line 127: OK, this makes a lot more sense now!

src/components/call/Call.tsx
```diff
@@ -87,62 +86,51 @@ export default class Call extends React.Component<Props, State> {
...
+});
+
+export default connect<
+  {
+    contactIndexes: Props['contactIndexes'];
```
nickoneill, line 132: Does typescript complain about missing types on the `contactIndexes` object here? It feels like we're leaving safety to address these string-ly

src/components/call/ContactProgress.tsx
```diff
@@ -195,3 +190,11 @@ export const ContactProgress: React.StatelessComponent<Props> = ({
     </div>
   );
 };
+
+const mapStateToProps = (state: ApplicationState) => ({
+  all: state.userStatsState.all
+});
+
+export default connect<{ all: UserContactEvent[] }, null, Omit<Props, 'all'>>(
```
nickoneill, line 198: Can you explain what `Omit` is here? I'm not sure I understand what it's purpose is.

nickoneill (commented):
OK, this is significantly simpler than what we have with context right now, I am onboard with doing things **the right way** with this pattern.

Can you share a bit about the renaming and what page / file structure that's going to look like with the whole repo converted too?

PR #494, opened 2019-02-22
Updated requirements and config for webpack 4

nickoneill:
Part of an effort to generate non-broken sourcemaps.

Webpack reports now that the js filesize is like 1.3mb, and then a more reasonable 361kb after gzip. I'm not sure if this is a regression but definitely something I need to look into before merging.

- [x] Check file size compared to previous version


nickoneill:
This seems to fix our sourcemap generation and the sizes have stayed the same (1.1mb before gzip, 366k after).

I'm going to open another issue to see if 1.1mb is including unintended stuff, it seems like a lot. But at least it isn't a change from current behavior.

PR #497, opened 2019-02-23
remove lodash, just import lodash-isequal to save bundle weight

aholachek:
Should save about 60kb minified/ungzipped.

We already polyfill find so I removed the lodash import for find.


src/components/call/CallPage.tsx
```diff
@@ -38,7 +37,7 @@ class CallPageView extends React.Component<Props> {
     }
     const currentIssueId = this.getIssueIdFromLocation(this.props);
     const previousIssueId = this.getIssueIdFromLocation(prevProps);
-    if (currentIssueId && !isEqual(currentIssueId, previousIssueId)) {
```
aholachek, line 41: no need for a special function to check string equality

nickoneill:
Thanks @aholachek! This looks great.

nickoneill:
Looks like we're finding on an undefined value at https://deploy-preview-497--5calls.netlify.com/more/

Shouldn't typescript give us an error about checking for possibly undefined values?

aholachek:
Yeah, I'm not sure, that page works for me when I run it locally, I guess I'll have to reimport lodash/find.
Bugsnag seems to swallow the stacktrace for me, I have to check the post request to see the error details, does that happen for you too?

aholachek:
I added a default prop of an empty array, it's working now-- it sort of makes sense that completedIssueIds can be null for some users

nickoneill:
I did not know about `defaultProps`! That's a very nice trick.

nickoneill (approved):

nickoneill:
Looks great, thank you!

PR #499, opened 2019-02-24
preload state

aholachek:


aholachek:
@nickoneill I figured out the problem here -- there are several places where we are putting non-serializable stuff into the redux store, e.g. [the methods on the Issue mode](https://github.com/aholachek/5calls/blob/5e8f88a866018f1702dcac65e2c49f5b6563a9a6/src/common/models/issue.ts#L31).

You [aren't supposed to put nonserializable state in redux](https://github.com/reduxjs/redux/blob/master/docs/faq/OrganizingState.md#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state) partially because it makes stuff like redux hydration impossible (since functions are not json serializable).

All functions on the models should probably go in some utility file that can be imported by the components as needed.

Any chance you can take a quick look at the various models and let me know how you'd like to refactor `models/contact.js`, `models/contactList.ts`, and `models/issue.ts` to not have methods? It seems to me that the ideal thing would be to get rid of the models altogether, but the fast hacky way would just be to ensure they are added back [as is done here](https://github.com/aholachek/5calls/blob/5e8f88a866018f1702dcac65e2c49f5b6563a9a6/src/redux/remoteData/reducer.ts#L36) but again, this sort of goes against the point of redux, which is supposed to be a serializable state tree.

nickoneill:
Contacts and the ContactList object aren't something we want to save since they're different for any given user. Can we exclude these from being serialized rather than trying to make them serializable? Rather, it seems like we want to only store some very general data (issue details, i.e. the currently selected call) for the detail pages, and not user state, not call completion, not the whole issue list, etc.

Can we do rehydration piecemeal with the stuff we *do* want rather than applying it to everything? It seems to me like there are a lot more items that we don't want to save than the stuff we do.

Also: Can we serialize custom classes without instance methods? Or do they have to go back to being interfaces (i.e. just objects) too?

aholachek:
Hi @nickoneill, I finally have it working.

I did it by instantiating the necessary instances in `remoteDataSate`. 
I still think a good long term goal would be to make redux just json in any case, this would go along nicely with #491 in the general move towards more idomatic usage of redux.

But I think the improvement in perceived loading time is dramatic enough that we should merge what I have here, assuming we hold off on adding new parts of remoteDataState that need methods as well (or as long as someone remembers to edit the hydrate store logic in that case).

re: "Also: Can we serialize custom classes without instance methods? Or do they have to go back to being interfaces (i.e. just objects) too?", I'm not sure I follow, what needs to be in redux besides just the objects? i can also go on slack if this should be more of a conversation.

nickoneill:
Code looks good, even though we're doing something that feels complex, it's easy to follow. The improvement in speed is super nice, very noticeable to get rid of that flash of no content we have now.

_One thing that should be fixed:_ If you check the deploy preview right now, it shows the issues from the last time the site was deployed. I checked the network panel and we're definitely not fetching the newest issues list on these pages.

I suspect this is because we've designed around local caching of issues for single session-length, i.e. we check if there are any issues and if there are, we skip fetching them. In this case there are always issues loaded since it's part of the preload cache.

A few operational details that are relevant to coming up with a solution:
* We frequently push new issues in the middle of the day and tweet them out immediately, so any solution needs to be able to pick up new issues immediately
* A flash of updated topics is probably OK (i.e. load with preload cache, then immediately fetch issues again to make sure we have the latest), this might be the simplest option but we should check the behavior of issues that are not yet in the preloaded cache but have just been made live, if only this specific case has a flash of "no issue here" it's OK
* In the future we'd like to be able to see expired issues on the site as well (issues that you should no longer call about) and the solution to this on the call page is probably going to look like: check for issue slug in loaded issues -> load issue details if it exists -> if not, request this specific issue from the server -> display issue if we get it back -> if not, display 404 page (just FYI, the solution might be related or might not)
* we can kick off nightly deploys to make sure the flashes of different content are minimal, but we should not rely on netlify deploys to deliver fresh content to users

nickoneill:
Hey @aholachek, I took a short break on this stuff but I'd love to get this merged in. Can I help coming up with a plan for what the rest of the work is on this?

aholachek:
Hi Nick! Sorry I kind of dropped the ball on this one. From what I understand I need to change my pr to always fetch new data, just do it in the background without wiping the current redux store, preventing  the visual jump but making sure users (eventually) get the most recent data. That makes sense to me. Then we'll have to do a bit of QA for the cases you mentioned.
I can probably get to this this weekend. I am happy to do it but also happy for you to start hacking on the pr if that's what you'd prefer.

nickoneill:
> From what I understand I need to change my pr to always fetch new data, just do it in the background without wiping the current redux store

That's pretty much it, yeah! That'll be enough to get it in at least, then we can make changes as needed.

PR #500, opened 2019-02-24
Added ie11 and edge to browser tests

nickoneill:


nickoneill:
😕 Refreshing this to take a look at IE / Edge compatibility again. We're not getting a bunch of errors on bugsnag, so I suspect this is quirks in how the tests are run in these browsers.

aholachek:
I know google analytics is not that super trustworthy, but according to them in the last week IE 11 and 9 were less than 1% of traffic, with average time spent on site of less than half that of all other major browsers (30 seconds).  To me this means focusing on, and even maybe supporting IE, is not a good use of time.
If we stop supporting IE11 I think in theory we could include fewer polyfills, transpile to a more up-to-date target, and theoretically save some bytes that way, also we could use new language features like grid. I don't have a strong opinion but it might be something to think about.

nickoneill:
I think that's fair, @aholachek. I'll update this PR to only add Edge and not test on IE 11 and we'll leave IE 11 and earlier out of our supported browsers in the future.

PR #507, opened 2019-02-28
WIP: Working on ways to fix snap only crawling a single page

nickoneill:
Let's try limiting concurrency to start.

### DIFF
--- a/package.json
+++ b/package.json
@@ -187,5 +187,8 @@
   "eslintConfig": {
     "extends": "react-app"
   },
+  "react-snap": {
+    "concurrency": 1
+  },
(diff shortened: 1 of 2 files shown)


aholachek (approved):
Seems worth a shot, although I can't say I understand how this might fix it

nickoneill:
The preview crawled OK, I'm going to check it if I change the JS too.

nickoneill:
Neither do I, but some suggestions in the snap issues list mentioned it.

Two successful deploys now, so I will merge later and then keep an eye on the next few master deploys.

PR #509, opened 2019-03-05
add blockquote styling, add meta description

aholachek:
if you run a lighthouse audit you see we are penalized for not having a standard `meta name="description"`, I don't know too much about seo but I don't think the meta itemprop="description" is a substitute.


nickoneill:
Not sure how we missed the normal meta description! Thanks, this looks great

nickoneill (approved):

PR #511, opened 2019-03-13
Fixes the react-snap deployment stuff finally

nickoneill:
...at least it does locally. We'll see what happens in netlify.

Webpack script links were all looking on 5calls.org for hashchunked urls (main.1234abcd.js) which didn't exist. A second deploy would always fix it because after the first deploy the js would then exist in prod.

### DIFF
--- a/package.json
+++ b/package.json
@@ -187,8 +187,6 @@
   "eslintConfig": {
     "extends": "react-app"
   },
-  "react-snap": {
-    "concurrency": 1
-  },
+  "react-snap": {},
   "postcss": {}
 }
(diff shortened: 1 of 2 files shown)


nickoneill:
YESSS, it worked the first time

package.json
```diff
@@ -187,8 +187,6 @@
   "eslintConfig": {
     "extends": "react-app"
   },
-  "react-snap": {
-    "concurrency": 1
-  },
+  "react-snap": {},
```
aholachek: should we just delete this

aholachek (approved):
Great find!! I can run react-snap locally on this branch, but I couldn't before.

package.json
nickoneill: Yeah, will remove unless we have another configuration line to add here

PR #574, opened 2020-03-23
Update loading experience

aholachek:
Before: (3 entire weird intermediary steps)

loading-before

After: (loading page)

loading-2

One thing that is worse than things being slow is things being confusing. It looks like the react snapped version that is being shown prior to page bootup is currently out of date.

I personally think one option would be to merge this pr that removes react-snap and invest more in speeding up the `issues` api endpoint if at all possible, as it currently takes more than a second on my fast internet connection, and that's the true blocker to showing users up to date information, here's a screenshot from the performance panel

Screen Shot 2020-03-22 at 9 28 40 PM

I'm still trying to figure out if I can speed things up, I wanted to make this pr so it would deploy on netlify and I could test it out.


nickoneill:
The timing for the issues request is interesting... I have datadog running on the API server and it reports that the responses take at most 20ms to generate and send back, and then on my end it looks like ~200ms to download. It's about 200kb in size so not exactly tiny. Can you run that request in curl and see if you still get the 1s+ response? Just a GET to `https://api.5calls.org/v1/issues` should be fine

nickoneill:
As for the snap changes, the static rendering seemed to improve our success rates for twitter, fb, etc picking up our `meta` tags with custom descriptions and images for each issue. Is there a benefit we can get from the improved loading states without having to remove snap?

My ideal here is actually to move entirely to static publishing and rebuild the issue list and call component as small, independent `preact` apps which should improve all of these concerns. If you look at the archive pages (https://5calls.org/archives/) they are entirely static HTML right now, but don't have any dynamic content on them.

aholachek:
Ah that makes sense, I'll close this. I would look into rendering the site
as a gatsby statically generated site which gets you both static pages +
automatic code splitting. I'll close this pr.

PR #576, opened 2020-03-26
Updated to our defer version of intercom for react

nickoneill:


nickoneill:
@aholachek Once this has a preview up, can you run your same test again? It looks like we already loaded intercom by `async` but I added `defer` as well to move script eval past the initial page load.

PR #581, opened 2020-05-22
Improved actblue embed donation

nickoneill:
Thanks to some guidance by @samuelcole, a much improved actblue embed experience without the extra state management in our app.

Still working on:
- [x] typescript types
- [x] done contributing experience

### DIFF
--- a/src/common/models/external.ts
+++ b/src/common/models/external.ts
@@ -1,5 +1,8 @@
 export interface ActBlue {
-  requestContribution(config: ActBlueConfig): Promise<ActBlueContribution>;
+  __initialized: boolean; // indicates that the actblue script is loaded and ready to go
+  requestContribution(
+    config: RequestActBlueContributionConfig
+  ): Promise<ActBlueContribution>;
   addEventListener(eventName: string, completion: ActBlueCompletionCallback);
 }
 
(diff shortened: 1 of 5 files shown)


src/common/models/external.ts
```diff
@@ -0,0 +1,32 @@
+export interface ActBlue {
+  requestContribution(config: ActBlueConfig): Promise<ActBlueContribution>;
+  addEventListener(eventName: string, completion: ActBlueCompletionCallback);
+}
+
+export type ActBlueCompletionCallback = (
+  contribution: ActBlueContribution
+) => void;
+
+export interface ActBlueConfig {
```
samuelcole: i would call this `RequestActBlueContributionConfig` since `actblueConfig` is a thing?

src/common/models/external.ts
```diff
@@ -0,0 +1,32 @@
+export interface ActBlue {
```
samuelcole, line 1: sorry, i'm not confident in my type definition syntax, so i'm not going to try:

`__initialized`: gets set by the actblue.js script, so you might be able to stub out the `actblue` interface before the script is loaded
`config`: is the derived configuration from `window.actblueConfig`, you might be able to dynamically change the config, by mutating stuff in here, but not advised
`dispatchEvent`: is the internal function used for dispatching events, might be useful for manually triggering callbacks?

src/common/models/external.ts
samuelcole:
```
      amount,
      amounts, // new option coming, could be string of comma seperated amounts in cents, or an array of cents in integers
      donor, // not donorData, but it does get renamed internally
      embedId, // you can specify your own id, which is echoed back in the `onContribute` events
      refcodes = {}, // actually an object of {refcodeBlah: 'blah'}
      styleSheetHref, // url to a stylesheet which will be injected in the form
      token
```

src/common/models/external.ts
```diff
@@ -0,0 +1,32 @@
...
+  lastname?: string;
+  email?: string;
+  zip?: string;
+}
+
+export interface ActBlueContribution {
+  amount: number;
+  embedId: string;
+  email?: string;
+  name: string;
```
samuelcole: this is confusing, it's the name of the _form_, so whatever slug you picked when you created the embed

src/common/models/external.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface ActBlueDonor {
+  firstname?: string;
+  lastname?: string;
+  email?: string;
+  zip?: string;
+}
+
+export interface ActBlueContribution {
```
samuelcole, line 30: missing:

```
recurring, // whether the contribution will recur at all
embedId, //automatically generated, or configured, embedId
refcodes // object containing all the refcodes for the contribution
```

src/common/models/external.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type ActBlueCompletionCallback = (
+  contribution: ActBlueContribution
+) => void;
+
+export interface ActBlueConfig {
+  token: string; // the config token for this embed, the only required attribute
+  amount?: string; // a pre-selected amount
+  donorData?: ActBlueDonor; // pre-entered information about the donor
+  express?: boolean; // maybe a bool indicating if express should be used?
```
samuelcole: this is used internally to trigger a form that processes the contribution immediately, via an express button, not externally exposed on `actblue.requestContribution`

samuelcole (commented):
just some quick notes on the type definitions

src/components/layout/Header.tsx
```diff
@@ -96,6 +104,23 @@ class HeaderImpl extends React.Component<Props, State> {
       });
   };
 
+  clickDonate(e: React.MouseEvent<HTMLAnchorElement>) {
+    if (window.actblue) {
```
samuelcole: this is where `__initialized` would be cool

samuelcole: Also it just occurred to me that you could send them off to an actblue.com page if `actblue` isn’t set, so that you could still capture the contribution if the script doesn’t load for whatever reason

nickoneill: I do! This click handler is on an href and I `preventDefault` if the actblue script is loaded, but if it's not then it defaults to the normal behavior and goes to the href.

src/common/models/external.ts
nickoneill: Thanks! These are helpful

src/common/models/external.ts
nickoneill: I think `__initialized` makes the most sense, expose too much of the internal workings and people just make a mess

nickoneill:
OK, updated the types with your notes @samuelcole. Feel free to grab them if you end up adding a react module

samuelcole (approved):
Love it! It was fun working with you today

PR #595, opened 2020-06-22
Issue 387 - progress circle on issues

### LINKED ISSUE #387: Add circular progress indicator for issue completion
Some folks can't call everyone at once, others only call a smaller selection of reps than what we send. It would be useful to provide a progress circle to let people know how far along they are if they haven't completed their reps.

Background in https://github.com/5calls/ios/issues/207

benjamincarp:
Fixes #387 

This PR adds progress to the circles on issue list items and it also makes it such that skipped contacts won't make an issue look incorrectly complete.

### DIFF
--- a/src/components/issues/IssuesList.tsx
+++ b/src/components/issues/IssuesList.tsx
@@ -13,11 +13,13 @@ interface Props {
   currentIssue?: Issue;
   completedIssueIds: string[];
   getIssuesIfNeeded: () => void;
+  getContactsIfNeeded: (force: boolean) => void;
 }
 
 export class IssuesList extends React.Component<Props> {
   componentDidMount() {
     this.props.getIssuesIfNeeded();
+    this.props.getContactsIfNeeded(false);
   }
 
   listFooter = () => {
--- a/src/components/issues/IssuesList.tsx
+++ b/src/components/issues/IssuesList.tsx
@@ -5,13 +5,14 @@ import { Link } from 'react-router-dom';
 import { Issue, ContactList } from '../../common/models';
 import { IssuesListItem } from './index';
 
-import { userStatsContext, remoteStateContext } from '../../contexts';
+import { userStatsContext } from '../../contexts';
 import { UserStatsState } from '../../redux/userStats';
 
 interface Props {
   issues: Issue[];
   currentIssue?: Issue;
   completedIssueIds: string[];
+  contacts: ContactList;
   getIssuesIfNeeded: () => void;
   getContactsIfNeeded: (force: boolean) => void;
 }
@@ -32,13 +33,14 @@ export class IssuesList extends React.Component<Props> {
     );
   };
 
-  listItems = (userStatsState: UserStatsState, contacts: ContactList) => {
+  listItems = (userStatsState: UserStatsState) => {
+    const { contacts, issues } = this.props;
     let currentIssueId = this.props.currentIssue
       ? this.props.currentIssue.id
       : 0;
 
-    if (this.props.issues && this.props.issues.map) {
-      return this.props.issues.map(issue => (
+    if (issues && issues.map) {
+      return issues.map(issue => (
         <IssuesListItem
           key={issue.id}
           issue={issue}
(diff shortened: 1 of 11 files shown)


src/components/issues/MoreIssues.tsx
```diff
@@ -106,24 +107,47 @@ export class MoreIssues extends React.Component<Props, State> {
...
-                      isIssueComplete={this.isIssueComplete(issue)}
-                      isIssueActive={false}
-                    />
-                  ))}
-                </ul>
-              </div>
-            ))
-          ) : (
-            <span />
+          <userStatsContext.Consumer>
```
nickoneill, line 111: Is there a place to include these at a higher level and pass user stats and remote state as props (or HOC them?), this is getting into pyramid of death levels of indentation.

src/components/issues/IssuesListItem.tsx
```diff
@@ -7,25 +7,27 @@ interface Props {
...
 }
 
 interface State {}
 
 export class IssuesListItem extends React.Component<Props, State> {
   render() {
-    const isCompleted = this.props.isIssueComplete ? 'is-complete' : '';
-    const isActive = this.props.isIssueActive ? 'is-active' : '';
-    const issueLink = `/issue/${this.props.issue.slugOrID()}`;
+    const { completeCount, contactsCount, isIssueActive, issue } = this.props;
```
nickoneill, line 18: Thank you for destructuring this, much cleaner

src/components/issues/IssuesList.tsx
```diff
@@ -55,10 +60,18 @@ export class IssuesList extends React.Component<Props> {
 
   render() {
     return (
-      <ul className="issues-list" role="navigation">
-        {this.listItems()}
-        {this.listFooter()}
-      </ul>
+      <userStatsContext.Consumer>
```
nickoneill, line 67: Similar comments as the other file in regards to two Consumers in a row adding indentation.

nickoneill (commented):
A couple small comments, but you're definitely on the right track. As far as I can tell the way you're calculating completed contacts should be sufficient for using to generate the completion circles.

src/components/issues/MoreIssues.tsx
benjamincarp: Remote state was loaded at a higher level so I passed it in through props and removed that level of context. I agree that the syntax for loading context is ugly but short of using Connect() or upgrading React to start using hooks I think this is the best we're going to get.

src/components/issues/IssuesListItem.tsx
benjamincarp: no problem.

src/components/issues/IssuesList.tsx
benjamincarp: Here too I was able to pass in the remote state context through props so it is a lot cleaner with only one layer of context here.

benjamincarp:
I ran into some issues with the test for IssueItem. It seems enzyme and context API are not friends. Every other instance of context in this repo appears to either have no unit test or uses snapshots. As you can see if you look at the snap files though, once it renders a context consumer it doesn't render any of the contents. 

I spent about an hour and a half last night just trying to get enzyme to render through the context consumer so I could get the test back to where it was but no luck. In the end, that test was really only checking if we were mapping over issues (issue item count matches issue count) so I don't think there is any real value loss in this change.

benjamincarp:
The end to end tests are breaking on master as well right (and would be out of scope for me to fix in this PR)?

nickoneill:
This looks great @benjamincarp! One note: when you initially load the page and you have no location set the circles are all filled in 100%, then adding a location sets them back to 0%. We should default them to 0% when we have no contacts yet.

The snapshot tests are not real useful, I would not miss them if they were removed or if consumers made them change too much. I know what the issue is with the e2e tests and I'll fix it outside of this change.

benjamincarp:
@nickoneill I had not been testing without location. Thanks for catching that. I put a fix in for that use case. 

Would you like me to just remove issuesList.test.tsx since neither remaining test is actually super useful or should I keep them there as a placeholder for someone to figure out how to make them really good later?

nickoneill:
Awesome, this looks great @benjamincarp. Ready for merge from your perspective?

Let's leave the test files as they are, I don't have a great plan for how to make them useful right now but perhaps someone will pop in with some more experience in this area and can easily update them.

benjamincarp:
Ready to merge!

nickoneill:
Thank you! This is going to make folks who call via the website very happy

